Issue #264, opened 2021-04-14
Convert a shelter to std::string


richelbilderbeek:
Currently, one cannot convert a shelter to a std::string. Converting to std::string is useful, for e.g. saving and debugging.

Make this test come true:

```c++
  //#define FIX_ISSUE_264
  #ifdef FIX_ISSUE_264
  // Conversion to string
  {
    const shelter s(1.2, 3.4, 5.6, color(7, 8, 9, 10));
    const std::string t = to_str(s);
    assert(!t.empty());
  }
  #endif // FIX_ISSUE_264
```

Do try to convert the shelter to an informative string, but we care less about your formatting: you are boss there! There is no `to_str` for color yet so either create that one (and test it!) or ignore that for now.

richelbilderbeek:
Hi @pohaha, well done!

Don't forget to, (1) remove the preprocessor statements, e.g.

```
  #define FIX_ISSUE_264
  #ifdef FIX_ISSUE_264
  // Conversion to string
  {
    const shelter s(1.2, 3.4, 5.6, color(7, 8, 9, 10));
    const std::string t = to_str(s);
    assert(!t.empty());
  }
  #endif // FIX_ISSUE_264
```

should be:

```
  // Conversion to string
  {
    const shelter s(1.2, 3.4, 5.6, color(7, 8, 9, 10));
    const std::string t = to_str(s);
    assert(!t.empty());
  }
```

Then, (2) Close the Issue (move to the 'Done' column does no close it)

These are details, but we are here to practice :-)

richelbilderbeek:
Hi @pohaha, I took the liberty to assign you and put the Issue in the 'To Do'. Will take only a couple of minutes to finish this Issue perfectly :+1:

PR #271, opened 2021-04-21
member method instead of a function?

pohaha:
so as per the task of issue #264 i've created necessary function in shelter.h/.cpp. but in my realization - it depends on another change (also mentioned in the issue) of color.h/.cpp i've added a (to_str) functionality to this module. But i made it as a class member of a color class instead of a separate function. should it be changed? 

also please tell if i'm doing a pr correctly in this case

### DIFF
--- a/color.cpp
+++ b/color.cpp
@@ -20,14 +20,6 @@ color::color(const int r, const int g, const int b, const int a)
     assert(m_a <= 255);
 }
 
-std::string color::to_str() const noexcept
-{
-    std::string msg = "["+std::to_string(m_r)+", ";
-    msg+=std::to_string(m_g)+", ";
-    msg+=std::to_string(m_b)+", ";
-    msg+=std::to_string(m_a)+"]";
-    return msg;
-}
 
 int get_blueness(const color &c) noexcept { return c.get_blue(); }
 
--- a/color.cpp
+++ b/color.cpp
@@ -37,6 +37,15 @@ int get_opaqueness(const color &c) noexcept { return c.get_opaqueness(); }
 
 int get_redness(const color &c) noexcept { return c.get_red(); }
 
+std::string to_str(const color& in_color) noexcept
+{
+    std::string msg = "[";
+    msg+=std::to_string(in_color.get_red())+", ";
+    msg+=std::to_string(in_color.get_green())+", ";
+    msg+=std::to_string(in_color.get_blue())+", ";
+    msg+=std::to_string(in_color.get_opaqueness())+"]";
+    return msg;
+}
 void test_color()
 {
     // Color is white by default
--- a/color.cpp
+++ b/color.cpp
@@ -20,6 +20,15 @@ color::color(const int r, const int g, const int b, const int a)
     assert(m_a <= 255);
 }
 
+std::string color::to_str() const noexcept
+{
+    std::string msg = "["+std::to_string(m_r)+", ";
+    msg+=std::to_string(m_g)+", ";
+    msg+=std::to_string(m_b)+", ";
+    msg+=std::to_string(m_a)+"]";
+    return msg;
+}
+
 int get_blueness(const color &c) noexcept { return c.get_blue(); }
 
 int get_greenness(const color &c) noexcept { return c.get_green(); }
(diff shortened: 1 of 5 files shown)


color.cpp
```diff
@@ -20,6 +20,15 @@ color::color(const int r, const int g, const int b, const int a)
...
 
+std::string color::to_str() const noexcept
+{
+    std::string msg = "["+std::to_string(m_r)+", ";
+    msg+=std::to_string(m_g)+", ";
+    msg+=std::to_string(m_b)+", ";
+    msg+=std::to_string(m_a)+"]";
+    return msg;
+}
+
```
richelbilderbeek, line 31: Nice touch with the square brackets and spaces after the comma!

game.pro
```diff
@@ -33,10 +33,10 @@ unix:!macx {
 }
 
 win32{
-  INCLUDEPATH += C:/Qt/sfml/include
-  INCLUDEPATH += D:/Qt/sfml/include
   LIBS += -LC:/Qt/sfml/lib
```
richelbilderbeek, line 36: These lines must be put back in, else the rest of the cannot run the game anymore. I know I ignored it while we were fixing your setup and I think this is the moment to fix that :-)

shelter.cpp
```diff
@@ -1,6 +1,7 @@
 #include "shelter.h"
 #include <cassert>
 #include <cmath>
+#include "string.h"
```
richelbilderbeek, line 5: The inclusion of an STL header file must be with with angle brackets [SF.12: Prefer the quoted form of #include for files relative to the including file and the angle bracket form everywhere else](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform). 

Either fix, or I will create an Issue to let a beginner fix this :+1:

shelter.h
```diff
@@ -3,6 +3,7 @@
 
 #include "color.h"
 #include <cmath>
+#include "string.h"
```
richelbilderbeek, line 6: The inclusion of an STL header file must be with with angle brackets [SF.12: Prefer the quoted form of #include for files relative to the including file and the angle bracket form everywhere else](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform). 

Either fix, or I will create an Issue to let a beginner fix this :+1:

game.pro
```diff
@@ -33,10 +33,10 @@ unix:!macx {
 }
 
 win32{
-  INCLUDEPATH += C:/Qt/sfml/include
-  INCLUDEPATH += D:/Qt/sfml/include
   LIBS += -LC:/Qt/sfml/lib
   LIBS += -LD:/Qt/sfml/lib
+  LIBS += -LC:/DEV/SFML-2.5.1/lib
+  INCLUDEPATH += /home/pohaha/Downloads/SFML-2.5.1/include
```
richelbilderbeek, line 39: This INCLUDEPATH clearly adds a Linux path, while being in the win32 section. I hope you will remove it, else I will create an Issue for a beginner to do so :-)

color.h
```diff
@@ -14,6 +14,7 @@ class color
   /// @param a alpha value or opaqueness (0 = transparant, 255 = opaque)
   color(const int r = 255, const int g = 255, const int b = 255,
         const int a = 255);
+  std::string to_str() const noexcept;
```
richelbilderbeek, line 17: I think I understand why you think this is a good idea: other languages usually do this. In C++, however, [C.4: Make a function a member only if it needs direct access to the representation of a class](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-member). Or, in proper English, if something can be done by a (free (as in non-member)) function, do so. For a color, we can already get the R, G, B and alpha value and that is all the info we need to convert a color to a string.

Either make this a free function, or I will create a beginner Issue to do so.

richelbilderbeek (changes requested):
Well done! I have given some suggestions to improve the architecture, with link to the C++ Core Guidelines. I would accept all these, as 'If all tests pass, we are -by definition- happy'. In `game.pro`, however, there is something that would break the build for other team members. I think our team members will love us even more if we fix that before merging to develop :-)
The other things are optional, but I hope you do these anyways. Else, no worries, I will make new Issues for those :-)

richelbilderbeek (approved):
Brilliant!

Issue #285, opened 2021-05-05
A game has a random number generator


swom:
The following test should pass in game.cpp:
```c++
#ifdef FIX_ISSUE_285
  {
    game g;
    int repeats = 1000;
    std::vector<double> numbers;
    int min = 0;
    int max = 100;
    std::uniform_real_distribution<double> unif_dist(min, max);
    double expected_mean = (max - min)/2;
    for(int i = 0; i != repeats; i++)
      {
        numbers.push_back(distr(g.get_rng()));
      }
    auto mean = calc_mean(numbers);
    assert(expected_mean - mean < 0.0001 && expected_mean - mean > -0.0001);
  }
#endif

```

richelbilderbeek:
Done!

PR #290, opened 2021-05-09
A game has a random number generator, #285

richelbilderbeek:
This was the results of a pair programming session of Pavel and I, hence quite some small commits. Also, there is some key-action-map code in it, which was the result of merging from develop to this branch I think :-/

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -63,6 +63,14 @@ void add_projectile(game &g, const projectile &p)
   g.get_projectiles().push_back(p);
 }
 
+double calc_mean(const std::vector<double>& v)
+{
+  return std::accumulate(
+    std::begin(v),
+    std::end(v), 0.0
+  ) / std::size(v);
+}
+
 double get_nth_player_size(const game& g, const int i)
 {
   return g.get_player(i).get_diameter();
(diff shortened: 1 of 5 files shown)


game.cpp
```diff
@@ -440,6 +448,11 @@ player game::wall_collision(player p)
   return p;
 }
 
+std::default_random_engine& game::get_rng() noexcept
```
TheoPannetier, line 451: I have a limited understanding of this but do we really want to pass the RNG by reference? Isn't there a risk it is overwritten through the game?

game.cpp
```diff
@@ -1145,11 +1168,12 @@ void test_game() //!OCLINT tests may be many
     std::uniform_real_distribution<double> unif_dist(min, max);
     double expected_mean = (max - min)/2;
     for(int i = 0; i != repeats; i++)
-      {
-        numbers.push_back(distr(g.get_rng()));
-      }
+    {
+      numbers.push_back(unif_dist(g.get_rng()));
```
TheoPannetier, line 1172: Very good name :)

game.cpp
TheoPannetier: Gotcha!

Issue #153, opened 2020-03-20
Prettier players


richelbilderbeek:
Currently, the players look kind of clunky. Go ahead and improve this!

There's a lot of liberty in this Issue, use it well :+1:

TheoPannetier:
What are things to keep in mind for the design? Any change to the colour or design of the players would make them less likely to camouflage on the background in #151, is that a feature we want to keep?

TheoPannetier:
depends on #151 imo

TheoPannetier:
**Resource:** https://www.sfml-dev.org/tutorials/2.5/graphics-shape.php

TheoPannetier:
I'm thinking of some kind of cute animal design for the players. Design should remain simple, or complex but always made of basic SFML shapes (`sf::CircleShape`, `sf::RectangleShape` etc), so it can still change colour based on the player's identity & state, and blend in the background.

TheoPannetier:
Idea for design: jumping spiders. Just like our players, jumping spiders move fast, are hungry, happily cannibalistic and love to ambush their preys. 

Don't think they can shoot rockets, but there are still species discovered every year so it who knows?

\+ they're *extremely* cute https://www.youtube.com/watch?v=GQP1MxxJt8I

TheoPannetier:
**Summary from the discussion on graphics**
Participants: @ollyturner @inesdaras @pohaha @TheoPannetier 

- Player's shape should be based on some animal, we picked a **dragon** 🐲 
- Player's shape should be drawn with some software externally from SFML (as an imported sprite), so the game doesn't have to generate an assemblage of shape by itself at every step
- Player's colour will be adjusted depending on the player's ID and state. We would have a base, fixed sprite e.g. [franjo.png](https://github.com/tresinformal/game/blob/master/franjo.png), then the colour would be changed by the game
- Player's colour scheme should be simple (mostly one colour?) so it can blend easily in coloured patches in the background (camouflage feature)

TheoPannetier:
There is an issue with the sprite not facing the same direction as the player, but I can't seem to fix it... I will just leave it there for now!

PR #292, opened 2021-05-12
Changed player sprite 🐉 

TheoPannetier:
#153
There is still an issue with the angle of the sprite (different from player), I can't manage to fix it. But players are ✨ prettier ✨


richelbilderbeek (approved):
I could not have done better! Well done finding out how the resources are handled!

PR #308, opened 2021-05-19
Fix #294

### LINKED ISSUE #294: A game_options has a seed for the random number generator
Currently, the random number generator (RNG ) seed is added slightly ad hoc. A more natural place would be within the game's options, similar to Minecraft's world creation seed.

Make the following test come true:

```c++
  #ifdef FIX_ISSUE_294
  // A game options has a seed for the random number generator
  {
    const game_options a;
    assert(a.get_rng_seed() == 0)
  }
  #endif // FIX_ISSUE_294
```

 * Do create a member variable for `game_options`, called `m_rng_seed`.
 * Do not worry about setting that seed yet, just use zero as the default value.

ollyturner:
pxArt

I and the the person in this image fixed issue #294!

### DIFF
--- a/game_options.cpp
+++ b/game_options.cpp
@@ -64,13 +64,11 @@ void test_game_options()
     music_on(o);
     assert(o.is_playing_music());
   }
-  #ifdef FIX_ISSUE_294
   // A game options has a seed for the random number generator
   {
     const game_options a;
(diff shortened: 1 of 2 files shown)


richelbilderbeek (approved):
Brilliant!

richelbilderbeek:
Also the art is awesome!

Issue #315, opened 2021-05-27
Spread shelters randomly over arena


richelbilderbeek:
Depends on:
- [x] #406
- [x] #408
- [x] #288 

Currently, shelters are a bit clumped up:

![Screenshot from 2021-05-27 07-06-50](https://user-images.githubusercontent.com/2098230/119769098-217d8200-beba-11eb-8da0-926fc26ec601.png)

Spead the shelter randomly over the arena using the game's RNG.

```cpp
// #define FIX_ISSUE_315
#ifdef FIX_ISSUE_315
  // Initial shelters are at random locations over the whole arena
  {
    // default game arguments
    double short_wall_side = 1600;
    int n_players = 0;
    int n_ticks = 0;
    int n_shelters = 10;
    int n_enemies = 0;
    int n_food = 0;

    int a_seed = 2;
    const game a_game(short_wall_side,
           n_players,
           n_ticks,
           n_shelters,
           n_enemies,
           n_food,
           a_seed
           );

    int another_seed = 3;
    const game another_game(short_wall_side,
           n_players,
           n_ticks,
           n_shelters,
           n_enemies,
           n_food,
           another_seed
           );

    const std::vector<coordinate> some_shelter_positions = get_all_shelter_positions(a_game);
    assert(some_shelter_positions.size() == n_shelters);
    const std::vector<coordinate> other_shelter_positions = get_all_shelter_positions(another_game);
    assert(other_shelter_positions.size() == n_shelters);

    assert(!all_positions_equal(some_shelter_positions, other_shelter_positions));
   }
#endif // FIX_ISSUE_315
```

janclod:
We made progress, test passes, but we need to clean up the code.

Also consider implementing a different solution to the problem. We currently build the shelter with default values (after doing a lot of math) and then randomize the position. We could simply get random coordinates in the first place and directly create the shelter using these random coordinates.

PR #316, opened 2021-05-27
Added more shelters

richelbilderbeek:
This was the code I wrote during group/mob programming last Wednesday, May 26th. 
I added tests, cleaned up a bit and now we now have 42 shelters.

Screenshot from 2021-05-27 07-06-50


game.h
```diff
@@ -22,7 +22,7 @@ class game
   game(double wall_short_side = 1600,
        int num_players = 3,
        int n_ticks = 0,
-       std::size_t n_shelters = 3,
+       std::size_t n_shelters = 42,
```
pohaha, line 25: Instead of setting a number of shelters across all tests manually, why don't we create a global variable(is this bad?) at the beginning of game.h file?

game_view.cpp
```diff
@@ -260,6 +261,17 @@ void game_view::show() noexcept
     }
 
 
+    if (1 == 2)
```
pohaha, line 264: Will this condition ever be reached?

game.cpp
```diff
@@ -890,19 +890,41 @@ void test_game() //!OCLINT tests may be many
...
+    const game g;
+    assert(g.get_shelters().size() == 42);
+  }
+  // There are 42 shelters
+  {
+    const int n_shelters{42};
+    const game g(1600, 3, 0, n_shelters);
+    assert(g.get_shelters().size() == n_shelters);
+  }
+  // All shelters have a different location
```
pohaha, line 904: Is this a look into future issue to create shapes on random places?

pohaha (commented):
Left some comments

pohaha:
Btw I do not know how to add my review comments to a comment like you did richel

game.h
richelbilderbeek: First, we should [I.2: Avoid non-const global variables](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-global). Second, there is now exactly one place where the default number of shelters is set, as a default argument, at the expected function (i.e. the constructor). I would say this is as good as it gets (although I know 1 equivalently good syntax)

game_view.cpp
richelbilderbeek: No, this is clearly to indivate 'This is a work in progress'. It was what I added at the session days after this Pull Request to show some code how to draw text for Theo. Would the review have been earlier (no need to apologize!) it would not have been.

game.cpp
richelbilderbeek: Yes, this is connected to #315 :-)

richelbilderbeek:
Hi @pohaha, 

I have replied to your questions. Thanks for checking up on me! What do I do next to get my review accepted :sunglasses:? 

Cheers, Richel

pohaha:
Hey quick question there. How do I accept only changes regarding shelters? Or is it not possible?

richelbilderbeek:
@pohaha you can ask me to undo the added ~~crap~~example code. And you should, as it will break develop and the team will suffer, so please reject for that reason :-)

richelbilderbeek:
@pohaha I fixed it now :-)

pohaha (approved):
Excellent richel as aloways!

PR #317, opened 2021-06-01
add class `read_only`

TheoPannetier:


richelbilderbeek (approved):
Brilliant work! All tests pass, so I am happy.

Some minor tidbits:

 * The project uses an indentation of 2 spaces in most code, changing this to 4 in main.cpp makes the coding style less coherent
 * Currently, `read_only` can or cannot have a value, as `optional` has. This is unneeded, `read_only` should always have a (read-only) value. There will be a new Issue to simplify `read_only`. BTW, don't forget, one can now make `optional<read_only<int>>` now!

Issue #261, opened 2021-04-07
[FD] It is possible to check if there is uneated food in the game


swom:
The following test should pass:
In game.cpp
```c++
void test_game()
{
 //Other tests

 #ifdef FIX_ISSUE_261
   {
  game g; //by default one uneaten food
    int n_food_items_begin = count_food_items(g);
    assert(has_food(g));
    eat_nth_food(g,0);
    assert(!has_food(g));
    //number of food item stays the same only the state of food item changes after they are eaten
    assert(n_food_items_begin == count_food_items)
   }
 #endif
}
```

PR #339, opened 2021-06-16
Issue #261

Clemcarle:


food.cpp
```diff
@@ -26,6 +26,15 @@ bool food::is_eaten() const noexcept {
...
 
+bool food::is_uneaten() const noexcept {
+    if (food::get_food_state() == food_state::uneaten) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+
```
TheoPannetier, line 37:
```suggestion
```
Nice that you added the complementary function for `food_state::uneaten`, but I feel it is unnecessary since strictly equivalent to `!is_eaten()`. To keep a good code coverage and limit the number of lines, I would remove it.

food.h
```diff
@@ -14,8 +14,11 @@ class food
   double get_y() const noexcept;
   double get_regeneration_time() const noexcept { return m_regeneration_time;}
   bool is_eaten() const noexcept;
+  bool is_uneaten() const noexcept;
```
TheoPannetier, line 17:
```suggestion
```

game.cpp
```diff
@@ -1066,126 +1200,150 @@ void test_game() //!OCLINT tests may be many
...
+    try {
+      game g; //by default one uneaten food
+      assert(has_food(g));
+      eat_nth_food(g,0);
+      assert(!has_food(g));
+      eat_nth_food(g,0); // throws exception
+    }
+    catch ( const std::exception& e ) {
+      assert(e == std::logic_error); // this is still wrong: make sure the error caught is the same as the error thrown
+    }
```
TheoPannetier, line 1224: I don't know what's our policy on try-catch statements. I feel this should be the role of an assert statement? I would need a second opinion on this - @richelbilderbeek ?

TheoPannetier (commented):
Good work @Clemcarle and @sebmader !
I have three comments I would like to see solved before approving the pull request:
1. I feel function `is_uneaten()` is unnecessary, and can be removed
2.    Indentation in `game.cpp` has been set to 4 spaces, while we agreed on using 2.
3. I see @sebmader added some try-catch. I don't think I have seen any other in the code, and I would like @richelbilderbeek 's opinion on their use vs asserts

richelbilderbeek:
@TheoPannetier thanks for the awesome report! 

> I see @sebmader added some try-catch. I don't think I have seen any other in the code, and I would like @richelbilderbeek 's opinion on their use vs asserts

Try-catches have their place, it is like `stop` in R; it allows a function to say 'Hey, I cannot do what you ask of me'. This is different than what an assert does ('I assume this is true, else I crash the program'). For more general functions, try-catch-throw definitely have their place!

TheoPannetier:
Alright, then I have no objection to the trycatches 👍

PR #344, opened 2021-06-30
Test validates that 'eat_nth_food' throws exception when food item is already eaten

sebmader:
We implemented a test with try-catch-block to make sure that exception is thrown.
We also omited the try-catch-block from the test that only checks for changing the food state because that should not thrown an exception.


game.cpp
```diff
@@ -1210,8 +1210,6 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
-  //#define FIX_ISSUE_340
-#ifdef FIX_ISSUE_340
   // make sure that eat_nth_food() throws a logic_error when the food is already eaten
   {
```
pohaha, line 1214: since this it the test the issue is required to operate.
The issue itself should be created and defined correctly, marked with [WIP]
this is needed to validate that the test is confirming the needed behavoir

game.cpp
```diff
@@ -1222,31 +1220,22 @@ void test_game() //!OCLINT tests may be many
       eat_nth_food(g,0); // throws exception
     }
     catch ( const std::exception& e ) {
-      assert(e.what() == "You cannot eat food that already has been eaten!");
+      assert(std::string(e.what()) == std::string("You cannot eat food that already has been eaten!"));
```
pohaha, line 1223: it's orbitrary, but this sort of test is testing if there are food to eat, and not testing the fact if the same food item is being eaten. Confusing

game.cpp
```diff
@@ -1222,31 +1220,22 @@ void test_game() //!OCLINT tests may be many
...
-    }
-    catch ( const std::logic_error& e ) {
-      std::cout << e.what();
-    }
+    game g; //by default one uneaten food
+    const int n_food_items_begin = count_food_items(g);
+    assert(has_food(g));
+    eat_nth_food(g,0);
+    assert(!has_food(g));
+    assert(n_food_items_begin == count_food_items(g));
```
pohaha, line 1235: this still needs to be defined in an issue (Other issue May be, since it is testing different behavioral patterns)

pohaha (commented):
There are a lot of things behind this commit. Though i saw the eat_nth_food function and desiphered it's meaning, it still should've been included in this commit, and i don't approve of

```c++
g.get_food()[n]
```
this behavoir the get_food should be invoked via index, and should return a food item, not the pointer to the whole array.

since this is not included in this PR, there is nothing i could do. But i also can't accept it.

pohaha (approved):
Even though the behavoir of the target tesk of the PR is completed, the inner structure is not a thing i can agree on. Though everything is fine now, this stuff should bee looked into later.
Approve for now

Issue #322, opened 2021-06-09
Create colors


richelbilderbeek:
The game uses colors to determine paper-rock-scissors dynamics. The increase the readability/Englishness of our code, we use the colors' names instead of their raw RGB values. For example, we use `create_red_color()` over `color(255,0,0)`. Create the three functions to create the three primary colors: 

```c++
  //#define FIX_ISSUE_322
  #ifdef FIX_ISSUE_322
  // Colors have the correct RGB values
  {
    const color red = create_red_color();
    assert(get_redness(red) == 255);
    assert(get_greenness(red) == 0);
    assert(get_blueness(red) == 0);
  }
  {
    const color green = create_green_color();
    assert(get_redness(green) == 0);
    assert(get_greenness(green) == 255);
    assert(get_blueness(green) == 0);
  }
  {
    const color blue = create_blue_color();
    assert(get_redness(blue) == 0);
    assert(get_greenness(blue) == 0);
    assert(get_blueness(blue) == 255);
  }
  #endif // FIX_ISSUE_322
```

TheoPannetier:
Oops, accidentally solved this as part of #229

TheoPannetier:
solved

Issue #229, opened 2021-03-24
[PRS][mathematics] 'calc_hue': convert RGB values to hue


richelbilderbeek:
The game has paper-rock-scissors dynamics, except it uses colors instead of paper, rock and scissors.
When two players touch, the color of the players determines the winner.
Instead of comparing RGB values (i.e. a 3-dimensional value) to determine the winner, one can use the color hue (i.e. a 1-dimensional value). Write the function `calc_hue` that calculates the hue of an RGB color.
Tip: use http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl/

```c++
  //#define FIX_ISSUE_229
  #ifdef FIX_ISSUE_229
  // Colors have the correct hues
  // See http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl/
  {
    const color red = create_red_color();
    assert(calc_hue(red) == 0.0);
  }
  {
    const color green = create_green_color();
    assert(calc_hue(red) == 120.0);
  }
  {
    const color blue = create_blue_color();
    assert(calc_hue(blue) == 240.0);
  }
  #endif // FIX_ISSUE_229
```

TheoPannetier:
solved

Issue #304, opened 2021-05-17
Write 'get_stun_key'


richelbilderbeek:
Currently, a player has a key-action-map (KAM) that connects a key to an action. For example, for player 1, the key `sf::Keyboard::Num1` is connected to `action::stun`, so that means that when player 1 presses '1', it will shoot a stun rocket. 

For #246 we need an easy way to get the key to shoot a stun rocket. Sure, we can get it from the KAM locally, but we can also make a helper function, called `get_stun_key`. Make the following test come true:

```c++
    #ifdef FIX_ISSUE_304
    //Get the stun key
    {
        const key_action_map m = get_player_1_kam();
        assert(get_stun_key(m) == sf::Keyboard::Num1);
    }
    #endif // FIX_ISSUE_304
```

**Protip:** read the comment thread for a shortcut 😉

TheoPannetier:
This has been (accidentally 😇) solved with #355, one can now call `m.to_key(action_type::shoot_stun_rocket)` to get the desired result.

That being said, defining a wrapper `get_stun_key()` around the line above would make a great `first` issue 🤔

ollyturner:
you were right solved

TheoPannetier:
Done!

PR #354, opened 2021-07-29
Random game options, fix #303, #322, #229, #355

### LINKED ISSUE #303: Write 'get_random_game_options'
* Needed for #246

The `game_options` class contains all the game options, i.e. those setting you normally set in an 'Options' screen. One set of options are the key-action maps (KAMs), as determined by the key configuration, i.e. which key corresponds to which action.

For testing, we need random `game_options`, especially with random KAMs, as this forces use to work with any random setup, e.g. we can check if a player actually moves forward when pressing its 'go forward' key.

Make the following tests pass:

```c++
  #ifdef FIX_ISSUE_303
  // Two random game options are the same, when the same seed is used
  {
    const int rng_seed = 271;
    const game_options a = get_random_game_options(rng_seed);
    const game_options b = get_random_game_options(rng_seed);
    assert(a == b);
  }
  // Two random game options differ in general
  {
    const int rng_seed = 314;
    const game_options a = get_random_game_options(rng_seed);
    const game_options b = get_random_game_options(rng_seed + 1);
    assert(a != b);
  }
  // Two random game options differ in their key-action maps especially
  {
    const game_options a = get_random_game_options(rng_seed);
    const game_options b = get_random_game_options(rng_seed + 1);
    assert(a.get_kam_1() != b.get_kam_1());
    assert(a.get_kam_2() != b.get_kam_2());
  }
  //#endif // FIX_ISSUE_303
```

The RNG seed is used to initialize a **local** RNG engine, such as a `std::default_random_engine` as used in the `game` class. No need to get fancy and re-use the `game`'s RNG engine.

TheoPannetier:
Note: the RNG used to draw random keys is passed to `game_options` class member `m_rng_seed`, but this is not used anywhere. In other words, changing the value of the `rng_seed` argument of the class constructor does not do anything. I'm unsure this behaviour is satisfying, but otherwise happy to merge.

EDIT: also made functions to create simple colors #322 
and made function `calc_hue()` #229

EDIT: now there's also a function to convert actions into keys 😀 #355


key_action_map.cpp
```diff
@@ -2,157 +2,195 @@
...
-                sf::Keyboard::RControl
-                );
+  for (auto& it : m_map)
+    {
+      if (it.second == action)
+        {
+          return it.first;
+        }
+    }
+  return sf::Keyboard::Key::Unknown;
```
richelbilderbeek, line 40: I like this constant! I didn't knew it existed!

color.cpp
```diff
@@ -29,62 +29,109 @@ int get_opaqueness(const color &c) noexcept { return c.get_opaqueness(); }
...
+    assert(get_opaqueness(c1)==10);
+  }
+  // operator==
+  {
+    const color c;
+    const color c1;
+    assert(c == c1);
+  }
+
+  //     operator!=
```
richelbilderbeek, line 117: No idea why so many spaces between `//` and `operator==`, no worries however :-)

color.h
```diff
@@ -47,6 +47,11 @@ int get_opaqueness(const color &c) noexcept;
 /// Get the redness
 int get_redness(const color &c) noexcept;
 
+/// Ready-made colors
```
richelbilderbeek, line 50: I enjoy the wording `Ready-made` colors :-)

game_options.cpp
```diff
@@ -43,7 +63,7 @@ void test_game_options()
   // Music can be switched off
   {
     const bool do_play_music = false;
-    const game_options g(do_play_music);
+    const game_options g(0, do_play_music);
```
richelbilderbeek, line 66: This edit scared me a bit, as the first impression would be that only the first player (with index 0) has no sound :-). I would have appreciated a named constant instead of the zero:

```c++
const int rng_seed = 0;
const game_options g(rng_seed, ...
```

Still, with minimal effort I could find out that zero is the RNG seed, so I will accept. I only want to share that named constants helps understand the code even better :-)

game_options.cpp
```diff
@@ -110,12 +131,27 @@ void test_game_options()
...
+  {
+    // Random game options should not draw keys already used by another player
+    const int rng_seed = 271; // by default this will assign A to both players
+    const game_options rgo = get_random_game_options(rng_seed);
+    key_action_map map_1 = rgo.get_kam_1();
+    key_action_map map_2 = rgo.get_kam_2();
+    bool a = map_1.has_key(sf::Keyboard::Key::A);
+    bool b = map_2.has_key(sf::Keyboard::Key::A);
+    // Key A is in KAM1 or KAM2 but not both
+    assert(a != b);
```
richelbilderbeek, line 152: This is a scary assert, chance 1 in 26^2 it will be triggered. This will happen one day. Let's keep it as-is, so a junior can 'debug' the problem when it happens :-)

key_action_map.cpp
```diff
@@ -2,157 +2,195 @@
...
+
+sf::Keyboard::Key get_random_key()
+{
+  return static_cast<sf::Keyboard::Key>(
+        static_cast<int>(sf::Keyboard::Key::A) + (std::rand() % 26)
+        );
+  // return static_cast<sf::Keyboard::Key>(std::rand() % 26);
+}
+
+key_action_map get_random_kam()
```
richelbilderbeek, line 98: Well done! This algorithm seems dauntingly powerful to me!

key_action_map.cpp
```diff
@@ -2,157 +2,195 @@
...
+    assert( m.has_key(sf::Keyboard::W));
+    assert( m.has_key(sf::Keyboard::S));
+    assert( m.has_key(sf::Keyboard::Q));
+    assert(!m.has_key(sf::Keyboard::L));
+  }
 
-    }
-  #define FIX_ISSUE_282
-  #ifdef FIX_ISSUE_282
+  //Giving a key that is not in the map returns a action_type::none action
```
richelbilderbeek, line 184: I understand the design choice. I do think one day, that this will throw an exception, but maybe not. Let's see what the mediors think :-)

game_options.cpp
```diff
@@ -110,12 +131,27 @@ void test_game_options()
   }
   // Two random game options differ in their key-action maps especially
   {
+    const int rng_seed = 111;
```
richelbilderbeek, line 134: Nice, adapting tests to facilitate new behavior! Note that this is what mediors do :innocent:

key_action_map.cpp
```diff
@@ -176,7 +214,51 @@ void test_key_action_map()//!OCLINT tests can be many
     assert(b != c);
   }
 
+#define FIX_ISSUE_282
+#ifdef FIX_ISSUE_282
+  {
+    const int rng_seed = 271;
+    std::srand(rng_seed);
```
richelbilderbeek, line 221: Well done, you've found `srand`!

key_action_map.h
```diff
@@ -21,12 +21,18 @@ class key_action_map
...
+  /// Find out which key triggers that action
+  sf::Keyboard::Key to_key(action_type action) const noexcept;
+
   ///Find out if the key is mapped
   bool has_key(sf::Keyboard::Key key) const noexcept;
 
   ///Returns const ref of m_map
   const std::map<sf::Keyboard::Key, action_type>& get_raw_map() const noexcept {return m_map;}
 
+  // Find the stun key
```
richelbilderbeek, line 33: This should one day be a free function, as we can already get this info from the `get_raw_map` member function result. I think I saw an Issue about that even already :-)

richelbilderbeek (approved):
Exemplary work! I've written down some future considerations and some compliments, which should be a fun read. Keep up the awesome work!

color.cpp
TheoPannetier: Not sure either, I didn't write these functions, just fixed the indentation I believe.
I will correct this in my next merge 👍

game_options.cpp
TheoPannetier: Definitely agree with this, I was unsure how to proceed as I'm used to R where you can just provide arguments as name-value pairs on a function call. I like the idea of declaring a variable just above for better readability!

game_options.cpp
TheoPannetier: I set `rng_seed = 271` because I know that drawing random keys with this seed assigns key A to both players 😉 
I just ran the program three times, the two key action maps were assigned the same values in all three iterations.
Or am I missing something about random number generations? Will the sequence be different from one computer to the next?

key_action_map.cpp
TheoPannetier: Oops, I left some commented code in there 😬 
I'm going to replace lines 92-94 with line 95, does the same job but in one line and with less conversions 🌈

key_action_map.cpp
TheoPannetier: Line 184 and the block that follows are not my work, I don't know what design choice has been made here.
I should stop touching the indentation/space of other people 😅

key_action_map.h
TheoPannetier: Ah yes good catch, this is a leftover of me addressing #304, then reverting my progress to let someone more junior resolve it. I forgot to remove this bit.
Now that we have `key_action_map::to_key` (see #355), solving #304 should be trivial. Should I make it a `first` issue?

game_options.cpp
richelbilderbeek: Ah, I guess I missed the `srand` somewhere earlier, my bad :blush:

key_action_map.h
richelbilderbeek: Making that a `first` Issue is great! A `first` Issue always has the tests written, as that is what makes it a `first`. I predict you'll happily write these as well :-)

game_options.cpp
TheoPannetier: `srand` is called inside `get_random_game_options` 😉

key_action_map.h
TheoPannetier: I put the test back 👍

Issue #256, opened 2021-04-07
[FD] A food that is eaten does not trigger collisons


swom:
Depends on
- [x] #261 
- [x] #269 
- [x] #247 

The following test should pass in game.cpp:
```c++
void test_game(){

//Other tests

#ifdef FIX_ISSUE_256

  {
    food f;
    player p;
    put_player_on_food(p, f);
    assert(in_contact_with_uneaten_food(p,f));
    eat_nth_food(g,0);
    assert(!in_contact_with_uneaten_food(p,f));
  }
#endif

}
```

TheoPannetier:
Done!

Issue #254, opened 2021-04-07
[FD] When a food is eaten its state is food_state::eaten 


swom:
Depends on 
- [x] #253
- [x] #236 
- [x] #260
- [x] #261 
The following test should pass:
In game.cpp:
```c++
#ifdef FIX_ISSUE_254
  {
    game g;
    put_player_on_food(g.get_player(0), g.get_food()[0]);
    g.tick();
    assert(is_eaten(g.get_food()[0]))
  }
#endif
```

TheoPannetier:
Done!

Issue #247, opened 2021-03-31
[FD] A player and uneaten food can be checked for collision


swom:
Depends on 
- [x] #237 
The following test should pass in game.cpp:
```c++

//Other tests

#ifdef FIX_ISSUE_247
  {
    player p;
    food f;
    assert(!in_contact_with_uneaten_food(p,f));
    put_player_on_food(p,f);
    assert(in_contact_with_uneaten_food(p,f));

  }
#endif
```

TheoPannetier:
Hold on @swom is this issue a duplicate of #237 ?
To me it looks like `have_same_position(player, food)` and `player_and_food_are_colliding` do the exact same thing.
Should I replace calls to the latter with the former?

TheoPannetier:
Aaaah, got it from #256, you want a function that checks that a player and an **uneaten** food item have the same position

TheoPannetier:
Done!

Issue #244, opened 2021-03-31
[FD]When player eats food, it becomes larger 


swom:
Depends on
- [x] #237 
- [x] #248 

The following test should pass in game.cpp

```c++

//Other tests

#ifdef FIX_ISSUE_244
  {
    game g;
    auto init_player_size = g.get_player(0).get_diameter();
    put_player_on_food(g.get_player(0), g.get_food()[0]);
    g.tick();
    assert(g.get_player(0).get_diameter() > init_player_size);

  }
#endif 
```

TheoPannetier:
Done!

PR #360, opened 2021-08-08
[FD] Players can now eat food and grow from it

TheoPannetier:
Fixed a bunch of food-related issues [FD]: #254 , #238 , #244, #247, #236, #256
I modified some of these issues and their tests to make it clearer what they do or request

Also:
- changed default food position to {2000, 2000} to avoid player-wall collisions issues when testing
- made `eat_nth_food()` a public member of `game` so that it can be called from within and access other members of the class
- removed function `has_enemy_collision` and the corresponding test, as this function wasn't checking anything at all and so the test was misleading

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -121,14 +121,6 @@ bool has_food(const game &g)
   return false ;
 }
 
-void game::eat_nth_food(const int n)
-{
-  if(get_food()[n].is_eaten()) {
-      throw std::logic_error("You cannot eat food that already has been eaten!");
-    }
-  get_food()[n].set_food_state(food_state::eaten);
-}
-
 int count_n_projectiles(const game &g) noexcept
 {
   return static_cast<int>(g.get_projectiles().size());
@@ -569,7 +561,7 @@ void game::make_players_eat_food()
        {
           if (are_colliding(player, get_food()[i]))
             {
-              eat_nth_food(i);
+              eat_food(get_food()[i]);
               player.grow();
             }
        }
@@ -582,6 +574,14 @@ void game::eat_food(food& f)
       throw std::logic_error("You cannot eat food that already has been eaten!");
     }
   f.set_food_state(food_state::eaten);
+}
+
+void eat_nth_food(game& g, const int n)
+{
+    if(g.get_food()[n].is_eaten()) {
+        throw std::logic_error("You cannot eat food that already has been eaten!");
+    }
+    g.eat_food(g.get_food()[n]);
 }
 
 std::default_random_engine& game::get_rng() noexcept
@@ -1300,10 +1300,10 @@ void test_game() //!OCLINT tests may be many
   {
     game g; //by default one uneaten food
     assert(has_food(g));
-    g.eat_nth_food(0);
+    eat_nth_food(g, 0);
     assert(!has_food(g));
     try {
-      g.eat_nth_food(0); // throws exception
+      eat_nth_food(g, 0); // throws exception
     }
     catch ( const std::exception& e ) {
       assert(std::string(e.what()) == std::string("You cannot eat food that already has been eaten!"));
@@ -1318,7 +1318,7 @@ void test_game() //!OCLINT tests may be many
     game g; //by default one uneaten food
     const int n_food_items_begin = count_food_items(g);
     assert(has_food(g));
-    g.eat_nth_food(0);
+    eat_nth_food(g, 0);
     assert(!has_food(g));
     assert(n_food_items_begin == count_food_items(g));
   }
@@ -1340,7 +1340,7 @@ void test_game() //!OCLINT tests may be many
     game g; //by default one uneaten food
     assert(has_food(g));
     auto initial_value_timer = get_nth_food_timer(g, 0);
-    g.eat_nth_food(0);
+    eat_nth_food(g, 0);
     assert(!has_food(g));
     g.tick();
     assert(init_value_timer + 1  == get_nth_food_timer(g, 0));
@@ -1350,7 +1350,7 @@ void test_game() //!OCLINT tests may be many
 #ifdef FIX_ISSUE_255
   {
     game g;
-    g.eat_nth_food(0);
+    eat_nth_food(g, 0);
     assert(nth_food_is_eaten(g,0));
     for(int i = 0; i != get_nth_food_regeneration_time(g, 0); i++)
       {
--- a/game.cpp
+++ b/game.cpp
@@ -570,7 +570,7 @@ void game::make_players_eat_food()
           if (in_contact_with_uneaten_food(player, get_food()[i]))
             {
               eat_nth_food(i);
-              // player.grow();
+              player.grow();
             }
        }
     }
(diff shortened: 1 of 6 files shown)


game.cpp
```diff
@@ -495,6 +479,34 @@ void put_player_on_food(player &p, const food &f)
...
+      p.get_y() - f.get_y() < 0.0001 &&
+      p.get_y() - f.get_y() > -0.0001;
+}
+
+bool in_contact_with_uneaten_food(const player &p, const food &f)
+{
+  return have_same_position(p, f) && !f.is_eaten();
+}
+
+bool any_player_food_collision(const game& g)
```
swom, line 495: Why not keep the naming for the 2 function functions 
```c++
has_food_collision(const game& g)
```
and 
```c++
are_colliding(const player& p, const food& f)
``` 
or 
```c++
have_collision(const player& p, const food& f)
```
? 
It is easier to read and conveys the message, especially when the test is read in #256. I find this nomenclature a bit clunky.

game.cpp
```diff
@@ -548,6 +560,30 @@ player game::wall_collision(player p)
...
+          if (in_contact_with_uneaten_food(player, get_food()[i]))
+            {
+              eat_nth_food(i);
+              player.grow();
+            }
+       }
+    }
+}
+
+void game::eat_food(food& f)
```
swom, line 579: I am not sure, but I feel there is some redundancy here between this function and 
```c++
game::eat_nth_food()
```
As C++ core guidelines suggest, we should keep member functions to a minimum and instead use as many free functions as possible. So I would suggest to make 
```c++
game::eat_nth_food()
```
a free function again
```c++
eat_nth_food(game& g, const int n)
```
that calls in turn the member function
```c++
game::eat_food(food& f) 
```
@richelbilderbeek what do you think? Is this line of reasoning correct or we can allowe both functions to exist as member functions?

game.cpp
```diff
@@ -1216,39 +1241,41 @@ void test_game() //!OCLINT tests may be many
...
 #ifdef FIX_ISSUE_247
   {
-    player p;
-    food f;
-    assert(!player_and_food_are_colliding(p,f));
+    double p_x = 12.3;
+    double f_x = p_x + 1.0;
+    player p{p_x};
+    food f{f_x};
+    assert(!in_contact_with_uneaten_food(p,f));
```
swom, line 1274: I think these lines are a bit unnecessary and make the test read less smoothly, while not conveying more meaning. I would revert back to the initial version.

game.cpp
```diff
@@ -1259,48 +1286,54 @@ void test_game() //!OCLINT tests may be many
...
 #ifdef FIX_ISSUE_256
   {
     food f;
     player p;
     put_player_on_food(p, f);
-    assert(player_and_food_are_colliding(p,f));
-    eat_nth_food(g,0);
-    assert(!player_and_food_are_colliding(p,f));
+    assert(in_contact_with_uneaten_food(p,f));
+    f.set_food_state(food_state::eaten);
```
swom, line 1335: Maybe wrap this in a free fucntion
```c++
gets_eaten(food& f)
```
just for the eyes.

game.cpp
```diff
@@ -301,6 +300,9 @@ void game::tick()
   //Check and resolve wall collisions
   do_wall_collisions();
 
+  // Make players eat food
+  make_players_eat_food();
```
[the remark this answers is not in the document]
TheoPannetier: Fair point! but I do prefer starting function names with an action verb whenever possible, by convention and for clarity. I could let myself convinced of this for a 5 or 6 words long name, but I feel 4 is still reasonable

game.cpp
richelbilderbeek: @swom free functions, when possible, are to be preferred. 

Note that `eat_food` probably finds the index of the food item, so it can call `eat_nth_food`  :-)

game.cpp
TheoPannetier: Yeaaah I do have to admit the names are a bit clunky indeed. But I choose them precisely bc I feel they better convey what the function does. 

`are_colliding(p, f)` → `in_contact_with_uneaten_food(p, f)`
Not clear what is meant that "colliding" means (same position + food uneaten). From the name I thought this just meant the two objects have the same position, e.g. this function was a duplicate of `have_same_position`.
So I went with a more explicit name.

`has_food_collision(g)` → `any_player_food_collision(g)`
I prefer player-food collision over food collision only (collision with what? other food items, enemies?)
I found that "any" makes it clearer that we're iterating across all players and food items in the game, but this isn't less important.
I can see now there is an issue of consistency, I used the word collision here but not in the function it calls.

All in all I think my issue is with a "collision" not being formally defined and being a not-so-obvious description of the situation (same position + able to interact). As a naive reader, "collision" just means same position to me. Though now I can see that collision refers to two objects interacting when they meet, as opposed to merely passing through each other (and I really can't think of a better name 🤔)

Happy to revert `in_contact_with_uneaten_food(p, f)` → `are_colliding(p, f)`, but then I think a definition of "collision" in inline comments would help reading the functions. I'll do that!

game.cpp
TheoPannetier: I don't like this situation either, and I did try to use `game::eat_nth_food()`, with no success. Can't recall why, I'll have a look again

game.cpp
TheoPannetier: I did this bc players and food items had the same default position of (0, 0) (and since food is uneaten, they were colliding).
Though this is no longer true, I think it's clearer to explicitly set them to different position. But I can collapse 1270-1273 into two lines.

game.cpp
TheoPannetier: That would make a great `first` issue! I'm calling dips on writing it 😛

game.cpp
richelbilderbeek: One day, there will be a member function `game::eat_food(std::iterator<food> iter)` (maybe called `remove_food` after a while), as using iterators is the preferred way to indicate an item in a vector. 

However, the team may not be ready yet to use iterators yet. But maybe there is a volunteer that wants to get more experience with algorithms and iterators.

For now, I feel using `game::eat_food`/`game::eat_nth_food` is a fine intermediary state.

game.cpp
TheoPannetier: I'm reverting changes to the names:

`any_player_food_collision(g)` → `has_any_player_food_collision(g)`
`in_contact_with_uneaten_food(p, f)` -> `are_colliding(p, f)`

All in all I think "collision" is a fairly good game, but I may have something more fitting to the situation: how about
`are_colliding(p, f)` -> `can_interact(p, f)`
`has_any_player_food_collision(g)` -> `has_any_player_food_interaction(g)` ?

game.cpp
TheoPannetier: Actually, this is exactly the job of `game::eat_food(f)`, although this is not a free function. I'll address this again after tackling the point above, when I get access to Qt again :)

game.cpp
TheoPannetier: Ok, as per 685f64f `eat_nth_food()` is a free function again and `game::eat_food()` is now a public function (so `eat_nth_food()` can call it) 😁

game.cpp
TheoPannetier: Ok, now I think we don't want this `gets_eaten()` function, as this would just be a free copy of `game::eat_food()`, and we would call it for just that one line

PR #363, opened 2021-08-25
Oliver

ollyturner:
Please review my work, @tzina97 and/or @kausmees. I am trying to make it so that a stun rocket will actually stun a rocket if it passes some if statements. Please tell me if you think there should be another check. Thanks

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -276,7 +276,7 @@ void game::projectile_collision()
                   this-> m_player[j].set_state(player_state::stunned);
 
                   // projectile disappears
-                  std::swap(m_projectiles[i], m_projectiles[m_projectiles.size()]);
+                  std::swap(m_projectiles[i], m_projectiles[m_projectiles.size()-1]);
                   this-> m_projectiles.pop_back();
 
                 }
--- a/game.cpp
+++ b/game.cpp
@@ -278,6 +278,16 @@ void game::projectile_collision()
                   std::swap(m_projectiles[i], m_projectiles[m_projectiles.size()-1]);
                   this-> m_projectiles.pop_back();
 
+                  // i can be invalid now, that is i can equal the number of projectiles
+                  // one option is:
+                  //
+                  //  --i;
+                  //
+                  // but this would be dangerous if the last project has disappeared
+                  //
+                  // a simple solution is:
+                  return;
+
               }
           }
       }
--- a/game.cpp
+++ b/game.cpp
@@ -268,10 +268,21 @@ void game::projectile_collision()
       }
       #endif // NEED_TO_WRITE_THIS_ISSUE_241
       // If the projectile touches the player ...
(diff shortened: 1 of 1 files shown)


kausmees:
Hey,

It says the checks have failed, but I'm not sure why that is since I can build and run your code.

Otherwise I think most things are good, the issue itself is not very clear so its hard to say if the code addresses everything.

However, I think there is a problem in the code in game::projectile_collision() on line 278-279

Using swap() and pop_back() to delete an item from a vector works, but it changes the order of elements in the vector (m_projectiles in this case.) Since this is done inside a for-loop that iterates over the indices in m_projectiles, it will cause a problem. 

For example, if the first projectile in the vector does hit a player, then the swap causes the last projectile in the vector to be swapped to the first place, and then the loop will never visit the last projectile (which is now in the first place in the vector).

I believe the vector.erase() method can be used to delete an element while preserving the order.

ollyturner:
Thats wierd.
I dont know why the code is failing.
I tried using vector.erase() but @richelbilderbeek told me not to and to use what I am using now. Maybe he didn't relalize. Let's ask him. I understand what you are saying so I will see what to do.

Thanks for reviewing my work.

richelbilderbeek:
I have fixed the access violation with some explanation. I wanted to show Oliver how problematic such an error is. I hope I showed a good example here :-)

game.cpp
```diff
@@ -251,10 +251,9 @@ void game::move_projectiles()
 void game::projectile_collision()
 {
 
-  const int n_projectiles = count_n_projectiles(*this);
   // For every projectile ...
 
-  for (int i = 0 ; i != n_projectiles ; ++i)
+  for (int i = 0 ; i != count_n_projectiles(*this) ; ++i)
```
richelbilderbeek, line 256: Brilliant!

game.cpp
```diff
@@ -1430,7 +1450,7 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
-  //#define FIX_ISSUE_241
+  #define FIX_ISSUE_241
```
richelbilderbeek, line 1453: This test passes, so the `#define`s and other useless fluff can be removed. I'll make that into a newbie Issue :-)

game.cpp
```diff
@@ -268,10 +267,31 @@ void game::projectile_collision()
...
+                  //  --i;
+                  //
+                  // but this would be dangerous if the last project has disappeared
+                  //
+                  // a simple solution is:
+                  return;
+
+              }
+          }
+      }
```
richelbilderbeek, line 293: Too complex, but that is the idea: I will make an Issue to simplify this, think `has_collision` :-)

richelbilderbeek (approved):
Definitely a step forward! Some cleaning needs to be done, however, such as removing unused code. Still, I will make that into newbie Issues :-)

ollyturner:
Thanks

Issue #231, opened 2021-03-24
[PRS] determine winner from two players


richelbilderbeek:
Depends on:

 * [ ] #230 

The game has paper-rock-scissors dynamics, except it uses colors instead of paper, rock and scissors.
When two players touch, there is always a winner following that dynamics.
The `is_first_player_winner` determines if the first player is the winner.

Make the following test come true:

```c++
  //#define FIX_ISSUE_231
  #ifdef FIX_ISSUE_231
  // The correct player must win
  {
    const player paper = create_red_player();
    const player rock = create_green_player();
    const player scissors = create_blue_player();
    assert(is_first_player_winner(paper, rock));
    assert(is_first_player_winner(scissors, paper));
    assert(is_first_player_winner(rock, scissors));
    assert(!is_first_player_winner(rock, paper));
    assert(!is_first_player_winner(paper, scissors));
    assert(!is_first_player_winner(scissors, rock));
  }
  #endif // FIX_ISSUE_231
```

Issue #230, opened 2021-03-24
[PRS] determine winner from two hues


richelbilderbeek:
Depends on:

 * [x] #229
 * [x] #322

The game has paper-rock-scissors dynamics, except it uses colors instead of paper, rock and scissors.
When two players touch, the color of the players determines the winner.
The `is_first_color_winner` determines if the first color is the winning color.
Use the `calc_hue` function to determine the winner.

Make the following test come true:

```c++
  //#define FIX_ISSUE_230
  #ifdef FIX_ISSUE_230
  // The correct color must win
  {
    const color paper = create_red_color();
    const color rock = create_green_color();
    const color scissors = create_blue_color();
    assert(is_first_color_winner(paper, rock));
    assert(is_first_color_winner(scissors, paper));
    assert(is_first_color_winner(rock, scissors));
    assert(!is_first_color_winner(rock, paper));
    assert(!is_first_color_winner(paper, scissors));
    assert(!is_first_color_winner(scissors, rock));
  }
  #endif // FIX_ISSUE_230
```

PR #371, opened 2021-09-15
#230 & #231

Clemcarle:
Issues #230 and #231 (paper-rock-scissors dynamics) have been addressed.

### DIFF
--- a/color.cpp
+++ b/color.cpp
@@ -87,6 +87,13 @@ std::string to_str(const color& in_color) noexcept
   return msg;
 }
 
+bool is_first_color_winner(const color& c1, const color& c2)
+{
+    double hue1 = calc_hue(c1);
+    double hue2 = calc_hue(c2);
+
+    return (hue1 == 0.0 && hue2 == 120.0) || (hue1 == 120.0 && hue2 == 240.0) || (hue1 == 240.0 && hue2 == 0.0);
+}
 
 void test_color()
 {
@@ -173,7 +180,7 @@ void test_color()
 #endif // FIX_ISSUE_229
 
 
-  //#define FIX_ISSUE_230
+#define FIX_ISSUE_230
 #ifdef FIX_ISSUE_230
   // The correct color must win
   {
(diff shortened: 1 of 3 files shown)


color.cpp
```diff
@@ -87,6 +87,13 @@ std::string to_str(const color& in_color) noexcept
   return msg;
 }
 
+bool is_first_color_winner(const color& c1, const color& c2)
+{
+    double hue1 = calc_hue(c1);
```
ollyturner, line 92: This is a bit of nitpicking, yet here goes: this should be const double, as hue_1 is not changed. Still, the tests pass, so I will make that into a newbie Issue 🙂

color.cpp
```diff
@@ -87,6 +87,13 @@ std::string to_str(const color& in_color) noexcept
   return msg;
 }
 
+bool is_first_color_winner(const color& c1, const color& c2)
+{
+    double hue1 = calc_hue(c1);
+    double hue2 = calc_hue(c2);
+
+    return (hue1 == 0.0 && hue2 == 120.0) || (hue1 == 120.0 && hue2 == 240.0) || (hue1 == 240.0 && hue2 == 0.0);
```
ollyturner, line 95: This was all that was needed to pass the test, well done.
Sidenote : there will be more tests 😸.

color.cpp
```diff
@@ -173,7 +180,7 @@ void test_color()
 #endif // FIX_ISSUE_229
 
 
-  //#define FIX_ISSUE_230
+#define FIX_ISSUE_230
```
ollyturner, line 183: This line can be romoved along with the other #ifstatements. @richelbilderbeek will make this cleaning up a newbie issue. 👍 :

ollyturner (approved):
Well done 🎉 You did an excellent job. Take a look at the comments, but no drastic changes needed.

richelbilderbeek:
Well done @kausmees and @Clemcarle ! Me and @ollyturner checked your handiwork!

Issue #389, opened 2021-09-29
[FD] A food item has a radius


TheoPannetier:
So far `food` objects are a single 1x1 square on the screen, but are drawn as if having a radius. Let's make a proper radius variable! 

```cpp
// A food has a radius member
  {
    const food f;
    assert(f.get_radius() >= 0.0);
  }
```

TheoPannetier:
Not yet on develop though

PR #390, opened 2021-09-29
expose #389

TheoPannetier:


richelbilderbeek (approved):
Brilliant, I could not have done better!

Issue #367, opened 2021-09-08
Player can move based on speed and direction[WIP]


swom:
Delete the following tests in `player.cpp` when assigning this issue

```
// A player can update its position
    {
        // TP: delete this test when assigning #367
        player p;
        // with initial position only x will change since sin of 0 is 0
        double a_x = p.get_x();
        double a_y = p.get_y();
        p.accelerate(); // move the player
        double b_x = p.get_x();
        double b_y = p.get_y();
        assert(std::abs(a_x - b_x) < 0.0000001);
        assert(std::abs(a_y - b_y) > 0.0000001);
    }
    {
        // TP: delete this test when assigning #367
        player p;
        p.turn_left();
        // change direction a little bit
        // otherwise default direction would not show
        // a change in y
        double a_x = p.get_x();
        double a_y = p.get_y();
        p.accelerate(); // move the player
        double b_x = p.get_x();
        double b_y = p.get_y();
        assert(std::abs(a_x - b_x) > 0.0000001);
        assert(std::abs(a_y - b_y) > 0.0000001);
    }
```

EvoLandEco:
After #525 , the tests were changed. Should be closed.

Issue #270, opened 2021-04-21
Make sure that backwards acceleration works properly


sebmader:
So far, going backwards is not implemented in game.
At least not practically but that will come at a later point in time.
Let's first make sure it theoretically works as it should by making some tests pass.
 
So, there is a function that makes the player accelerate backwards, but it does so by immediately setting the players speed to the maximum possible speed.
But - as it is implemented with forward acceleration - we want acceleration to work incrementally: you get faster until you reach maximum speed.

Make the following tests pass:

```
 // A player increases its backward speed by one 'backward acceleration' per backward acceleration
  // or: a player decreases its speed by one 'backward acceleration' per backward acceleration
  {
      player p;
      p.acc_backward();
      assert(std::abs(p.get_speed() - p.get_acceleration_backward()) < 0.00000000001);
  }
  // A players speed after one 'acceleration' is less than max_speed
  {
      player p;
      p.accelerate();
      assert(p.get_speed() < p.get_max_speed());
  }
  // A players speed after one 'backward acceleration' is more than negative max_speed
  {
      player p;
      p.acc_backward();
      assert(p.get_speed() > -p.get_max_speed());
  }
  // RJCB: my suggested test
  // A players goes ?right/?up upon acceleraton
  {
      player p_forward;
      p_forward.accelerate();
      assert(get_dx(p_forward) > 0.0); // Get the delta x, maybe needs to be added
      assert(get_dy(p_forward) > 0.0); // Get the delta y, maybe needs to be added
      player p_backward;
      p_backward.acc_backward();
      assert(get_dx(p_backward) < 0.0); // Signs should flip
      assert(get_dy(p_backward) < 0.0); // Signs should flip
  }
  // RJCB: another suggested test
  // A players actually goes backwards after some backwards accelerations
  {
      player p;
      p.accelerate();
      assert(get_dx(p) > 0.0); // Get the delta x, maybe needs to be added
      assert(get_dy(p) > 0.0); // Get the delta y, maybe needs to be added
      p.acc_backward();
      p.acc_backward();
      p.acc_backward();
      assert(get_dx(p) < 0.0); // Signs should flip
      assert(get_dy(p) < 0.0); // Signs should flip
  }
#endif
```

You will have to add the functions `get_dx()` and `get_dy()` which are meant as a measure of speed:
How much distance is covered per tick?
Additionally, they are free function that ought to be implemented in the player files but outside of the class.

EvoLandEco:
Closed by #525

PR #391, opened 2021-09-29
Sebastian

sebmader:
We updated the tests for #270.
They overlap quite a bit with the ones of #367 that we discussed today but I feel, it makes sense to stick with the original plan of implementing backwards acceleration the same way as forward acceleration is implement right now, and change it together with the split of acceleration and move.


sebmader:
I don't know where all these changes are coming from....?!
Will look at it next time..

richelbilderbeek:
@sebmader I guess some changes come from the merge of develop to the sebastian branch. I guess the review only needs to focus on the acceleration changes, on not the other changes that have reached develop :-)

TheoPannetier:
Nope, let's not merge from `sebastian` to `master` %)

player.cpp
```diff
@@ -584,26 +590,45 @@ void test_player() //!OCLINT tests may be long
     // A players goes ?right/?up upon acceleraton
     {
         player p_forward;
+        coordinate c_before = get_coordinate(p_forward);
         p_forward.accelerate();
```
TheoPannetier, line 594: A bit confued here: haven't we agreed that `accelerate()` should *not* move the player?

player.cpp
```diff
@@ -584,26 +590,45 @@ void test_player() //!OCLINT tests may be long
...
         player p;
+        coordinate c_before = get_coordinate(p);
         p.accelerate();
-        assert(get_dx(p) > 0.0); // Get the delta x, maybe needs to be added
-        assert(get_dy(p) > 0.0); // Get the delta x, maybe needs to be added
+        coordinate c_inbetween = get_coordinate(p);
+        double dx_a = get_x(c_inbetween) - get_x(c_before);
+        double dy_a = get_y(c_inbetween) - get_y(c_before);
+        assert(dx_a > 0.0); // Get the delta x
+        assert(dy_a > -0.00000000001 && dy_a < 0.00000000001);
```
TheoPannetier, line 622: That bit is redundant with the block just above isn't it?

player.cpp
```diff
@@ -584,26 +590,45 @@ void test_player() //!OCLINT tests may be long
...
         p.acc_backward();
         p.acc_backward();
-        assert(get_dx(p) < 0.0); // Signs should flip
-        assert(get_dy(p) < 0.0); // Signs should flip
+        coordinate c_after = get_coordinate(p);
+        double dx_b = get_x(c_after) - get_x(c_inbetween);
+        double dy_b = get_y(c_after) - get_y(c_inbetween);
+        assert(dx_b < 0.0); // Signs should flip
+        // it should not change in the y direction if the assumption about initial direction is correct
+        assert(dy_b > -0.00000000001 && dy_b < 0.00000000001);
```
TheoPannetier, line 631: Oh that's brilliant, I like this being checked for

player.cpp
sebmader: What do you mean? Pls be more specific :)

PR #398, opened 2021-10-13
The players have different keys for the "stun rocket" action

kausmees:
Changed the player keymaps for shooting stun rockets as follows:

player 1: E
player 2: O
player 3: Return

We're not sure about Return being a good option, but due to the fact that we have different keyboards, couldn't think of a better one.

### DIFF
--- a/game_view.cpp
+++ b/game_view.cpp
@@ -158,7 +158,7 @@ void game_view::draw_food() noexcept
 
 void game_view::press_key(const sf::Keyboard::Key& k)
 {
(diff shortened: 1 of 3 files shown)


TheoPannetier:
Great work @kausmees and @ollyturner 🚀 
My only objection is about using the Return key indeed.
Wouldn't it make sense to change keys for regular and stunning rockets of player 3 from Ctrl and Return to keys 7 and 9 respectively? AFAICS these keys are not yet in use.
Happy to merge regardless of your decision on this point!

PR #418, opened 2021-11-03
Increased CodeCov

ollyturner:


TheoPannetier:
Approved, those are small changes and they don't break the build 👍

Issue #400, opened 2021-10-20
A game's min and max coordinates can be accessed quickly


TheoPannetier:
The game's min and max x and y can currently be accessed through `g.get_env().get_max/min_x/y());`, which is a bit clunky to read and write. It would be nice to have wrapper functions for these.

```cpp
// #define FIX_ISSUE_400
#ifdef FIX_ISSUE_400
  // A game's min and max coordinates can be accessed quickly
  {
    game g;
    double max_x = get_max_x(g);
    assert(max_x == g.get_env().get_max_x());
    double min_x = get_min_x(g);
    assert(min_x == g.get_env().get_min_x());
    double max_y = get_max_y(g);
    assert(max_y == g.get_env().get_max_y());
    double min_y = get_min_y(g);
    assert(min_y == g.get_env().get_min_y());
  }
#endif
```

Issue #250, opened 2021-04-07
[FD] Food items can be randomly placed in the game 


swom:
Depends on:
- [x] #285 
- [x] #400 

The following test should pass in game.cpp:
```c++
// #define FIX_ISSUE_250
#ifdef FIX_ISSUE_250
  //Food can be placed at a random location
  {

    game g;
    std::vector<double> food_x;
    std::vector<double> food_y;

    int repeats = 1000;

    for(int i = 0; i != repeats; i++)
      {
        place_nth_food_randomly(g,0);
        food_x.push_back(get_nth_food_x(g,0));
        food_y.push_back(get_nth_food_y(g,0));
      }
    auto mean_x = calc_mean(food_x);
    auto mean_y = calc_mean(food_y);

    // Mean position of food items should be center of game environment
    double expected_mean_x = (get_max_x(g) + get_min_x(g)) / 2.0;
    double expected_mean_y = (get_max_y(g) + get_min_y(g)) / 2.0;
    double d_mean_x = abs(expected_mean_x - mean_x);
    double d_mean_y = abs(expected_mean_y - mean_y);

    assert(d_mean_x < 0.01);
    assert(d_mean_x < 0.01);

    auto var_x = calc_var(food_x, mean_x);
    auto var_y = calc_var(food_y, mean_y);

    assert(var_x > 0.01);
    assert(var_y > 0.01);
  }
#endif
```

TheoPannetier:
Hey @swom I'm probably ;issing something but shouldn't the conditions in assert be different?
Mean coord == 0 and var coord == 0 can only be obtained with a vector of food items that all have position (0,0). 
Doesn't sound like random placement to me 😉

richelbilderbeek:
I unassiged Gina, as she left the team for now :-)

TheoPannetier:
@kausmees I don't think the test needs to be changed. From what I can see the numbers sampled by `place_randomly()` are always the same, maybe because the RNG is generated inside the function and so reset at every call.
I tried passing the game's seed (through `g.get_rng()`) but that did not fix the issue...

richelbilderbeek:
Me and Theo simplified the test quite a lot! Good stuff!

TheoPannetier:
Done 😎

PR #419, opened 2021-11-03
Food can be placed randomly

TheoPannetier:
- #250 
- #400

### DIFF
--- a/food.cpp
+++ b/food.cpp
@@ -4,6 +4,7 @@
 #include <cmath>
 #include <sstream>
 
+
 food::food(const coordinate c, const color &col, const int regeneration_time, food_state food_state, double radius)
   : m_c{c}, m_color{col}, m_regeneration_time{regeneration_time}, m_food_state{food_state}, m_radius{radius}
 {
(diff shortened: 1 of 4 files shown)


food.cpp
```diff
@@ -33,6 +34,14 @@ bool operator==(const food& lhs, const food& rhs) noexcept
             && lhs.get_y() == rhs.get_y() ;
 }
 
+void food::place_randomly(std::mt19937 &rng, const double min_x, const double max_x, const double min_y, const double max_y){
```
swom, line 37: We could even give this function two `coordinate`  arguments
```suggestion
void food::place_randomly(std::mt19937 &rng, const coordinate& top_right, const coordinate& bottom_left){

```
And then deduce the `x`s and `y`s from those in the body of the fucntion

game.cpp
```diff
@@ -76,6 +76,32 @@ void add_projectile(game &g, const projectile &p)
...
+double get_min_x(const game &g){
+    return g.get_env().get_min_x();
+}
+
+double get_max_y(const game &g){
+    return g.get_env().get_max_y();
+}
+double get_min_y(const game &g){
+    return g.get_env().get_min_y();
+}
```
swom, line 93: And to link to the previous comments we can write free functions:
```c++
coordinate get_top_right_corner(const game& g)
coordinate get_bottom_left_corner(const game& g)
```

swom (commented):
Well done! I left a couple of comments on how to make this even fancier! 🤺

Issue #383, opened 2021-09-22
'game_options' must have an 'environment_type'


richelbilderbeek:
Currently, we do have some cool `environment_type`s (e.g. `attractive`, `repellent` and `wormhole`), but there is no way to get in into the game yet.

First step is to add a member variable to hold an `environment_type` into `game_options`.

Write a test that fails, then make it come true :-)

TheoPannetier:
Done!

Issue #370, opened 2021-09-15
Define 'get_stun_key(game_view&)


richelbilderbeek:
From [here](https://github.com/tresinformal/game/issues/304#issuecomment-889434640):

> This has been (accidentally innocent) solved with #355, one can now call 
> m.to_key(action_type::shoot_stun_rocket) to get the desired result.
>
> That being said, defining a wrapper get_stun_key() around the line above would
> make a great first issue thinking

PR #421, opened 2021-11-10
Oliver

TheoPannetier:
- #304 
- #370
- #383 

@richelbilderbeek we changed the test in #304 to call a free function instead of a member of `key_action_map`, and removed #370 which was redundant and had weird out-of-place definition of `get_stun_key`. Good moves?


game_options.cpp
```diff
@@ -159,5 +162,14 @@ void test_game_options()
   }
   #endif // FIX_ISSUE_353
 
+ #define FIX_ISSUE_383
+#ifdef FIX_ISSUE_383
+// A game_options has an environment_type member
+{
+  game_options go;
```
richelbilderbeek, line 169: The be very precise, this should be `const game_options go`, as reading/getting data should/does not modify the `game_options`

richelbilderbeek (approved):
Excellent work, could not have done better, except for one `const` that is missing. 

Let's merge it!

Issue #262, opened 2021-04-14
Functions must return `std::string` instead of `const std::string`


richelbilderbeek:
Since move semantics, returning a `const std::string` is recommended against, according to [F.15: Prefer simple and conventional ways of passing information](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f15-prefer-simple-and-conventional-ways-of-passing-information) where the return type is `X` (and not `const X`).

For example:

```c++
const std::string to_str(action_type this_action_type);
```

Should nowadays be:

```c++
std::string to_str(action_type this_action_type);
```

Look in the code for `const std::string` being returned and fix that. This needs to be done in both the .h file and the .cpp file.

richelbilderbeek:
I re-open, as I found another example. Good lucking finding it :+1:

richelbilderbeek:
I have unassigned Xiaoyan, as she has been kept occupied by work for too long now :-). If she gets back to us one day, she can assign herself again :-)

richelbilderbeek:
As of today, there is one in `shelter.h`. Please fix :rainbow:

ollyturner:
sure

kausmees:
Has been done.

PR #423, opened 2021-11-10
Fixed issue #262

decommandpro:


shelter.h
```diff
@@ -48,6 +48,6 @@ int get_opaqueness(const shelter &f) noexcept;
 
 /// Get the redness (from the color) of the shelter
 int get_redness(const shelter &f) noexcept;
-const std::string to_str(const shelter& in_shelter) noexcept;
+std::string to_str(const shelter& in_shelter) noexcept;
```
swom, line 51: Immagine this line is wrong...
```suggestion
std::string  to_str(const shelter& in_shelter) noexcept;
```

swom (changes requested):
Great but imagine I want you to change something

Issue #341, opened 2021-06-16
food equality not working correctly


pohaha:
as of now the equality operation can be performed on food objects
but it does not work as intended:
foods that are on the same coordinate point but with different colour ARE equal, when this is not true in practice.
to fix this issue the following test should pass:
```c++
#ifdef FIX_ISSUE_341
    {
      const food test_food_one(0.0f,0.0f,color());
      const food test_food_two(0.0f,0.0f,color(0,0,0));
      assert(not (test_food_one==test_food_two));
    }
```

Yangshen0325:
We confirmed it is working. We close the issue now.

Issue #257, opened 2021-04-07
[FD] A food that returns to the food_state::uneated state is relocated in a random position


swom:
- [x] #250 
- [x] #403
- [x] #404 

When a food item regenerates after being eaten, it should re-appear at a different, random position.
Make sure to use `place_nth_food_randomly()` from #250 or a similar function :wink:

```cpp
// #define FIX_ISSUE_257
#ifdef FIX_ISSUE_257
  {
    // A food that returns to the uneaten state is relocated at random
    game g;
    assert(!is_nth_food_eaten(g, 0));
    coordinate food_pos_before = get_nth_food_position(g, 0);
    eat_nth_food(g, 0);
    // Tick until food regens
    while (is_nth_food_eaten(g, 0)) {
      g.tick();
    }
    coordinate food_pos_after = get_nth_food_position(g, 0);
    assert(food_pos_after != food_pos_before);
  }
#endif
```

TheoPannetier:
Test is minimal so far because it doesn't assert that the new position is random, or that it will be different every time the food regenerates.

Issue #411, opened 2021-10-27
All classes that have a position now get that position from free logistic functions and not member functions.


swom:
For classes:
- [ ] #416
- [ ] #415
- [ ] #414
- [ ] #413
- [ ] #412

Change all calls to member functions that read the  `m_x `or `m_y `member, i.e.:
```c++
a_class.get_x(); //or
a_class.get_y();
```
to free logistic functions, i.e.:
 ```c++
get_x(const a_class& a_class);
get_y(const a_class& a_class);
```
do this  in both tests and the code. 
The only functions allowed to call member functions to read x and y 
```c++
a_class.get_x(); //or
a_class.get_y();
```
are the free logistic functions you will have used everywhere else
 ```c++
get_x(const a_class& a_class);
get_y(const a_class& a_class);
```

### **To be noted**:

1. If a free_logistic function with the same name exists for the class **_DO NOT_** write a new one, or change the existing one, just swap it with calls to the member function.
2. you **_SHALL NOT_** delete the definition and declaration of 
```c++
a_class.get_x(); //or
a_class.get_y();
```
from a class, keep it, just do not let anybody except the free logistic functions use it.

TheoPannetier:
Would a template be a good idea here? I'm tempted to take advantage of the 'get_position()' function that all items share and do it all at once with 

```cpp
template <typename T>
double get_x(const T& item)
{ 
    return item.get_position.get_x();
}
template <typename T>
double  get_y(const T& item)
{ 
    return item.get_position.get_y();
}
```

or is some good ol' function overloading preferable?

richelbilderbeek:
@TheoPannetier: this is a fun idea and I am happy you found it (and I could not help fix the code).  

I would not mind gradually adding this to the project. It would be a good experience for some team members to experiment with templates. 

Argument against doing so, is that no new functionality is gained (just existing one removed) and the code will be harder for newbies to understand. 

I think with the current state of the team, I think the pros outweight the cons :-)

```c++
template <typename T>
double get_position(const T& item)
{ 
    return item.get_position();
}
template <typename T>
double get_x(const T& item)
{ 
    return get_position(item).get_x();
}
template <typename T>
double  get_y(const T& item)
{ 
    return get_position(item).get_y();
}
```

PR #425, opened 2021-11-14
solving #257 #250 #341

pohaha:
we have fixed the problem with food not being placed randomly. and all tests that broe due to it/ It was done on your branch - so you shold at leat see if it all works with you^) gl.


swom (commented):
There is a way to improve the test for #257, probably @Clemcarle can help. Basically just stitch a for loop around lines `1717` up to `1723` and repeat it a few thousand times, collect the `coordinate`s generated inside the loop in a `vector` and then check that those `coordinate`s' `x` and `y` have mean and variance corresponding to the mean and variance of an uniform distribution with the same ranges of `x` and `y`. 
@richelbilderbeek do you think there is another more elegant way to prove that they are coming from a distribution?

food.cpp
```diff
@@ -29,9 +33,23 @@ bool food::is_eaten() const noexcept {
 
 bool operator==(const food& lhs, const food& rhs) noexcept
 {
-    return lhs.get_x() == rhs.get_x()
-            && lhs.get_y() == rhs.get_y()
-            && lhs.get_color() == rhs.get_color();
+    bool is_color_equal = (lhs.get_color() == rhs.get_color());
```
TheoPannetier, line 36:
```suggestion
    bool is_color_equal = lhs.get_color() == rhs.get_color();

```

food.cpp
```diff
@@ -29,9 +33,23 @@ bool food::is_eaten() const noexcept {
 
 bool operator==(const food& lhs, const food& rhs) noexcept
 {
-    return lhs.get_x() == rhs.get_x()
-            && lhs.get_y() == rhs.get_y()
-            && lhs.get_color() == rhs.get_color();
+    bool is_color_equal = (lhs.get_color() == rhs.get_color());
+    return get_x(lhs) == get_x(rhs) &&
+          (get_y(lhs) == get_y(rhs)) &&
```
TheoPannetier, line 38:
```suggestion
          get_y(lhs) == get_y(rhs) &&

```

coordinate.cpp
```diff
@@ -1,4 +1,5 @@
 #include "coordinate.h"
+#include "game.h"
```
TheoPannetier, line 2:
```suggestion
```

enemy.cpp
```diff
@@ -1,4 +1,5 @@
 #include "enemy.h"
+#include "game.h"
```
TheoPannetier, line 2:
```suggestion

```

food.cpp
```diff
@@ -3,6 +3,10 @@
 #include <cassert>
 #include <cmath>
 #include <sstream>
+#include <stdio.h>      /* printf, scanf, puts, NULL */
+#include <stdlib.h>     /* srand, rand */
+#include <time.h>
+#include <iostream>
```
TheoPannetier, line 9:
```suggestion
```

TheoPannetier: I dont't think these libraries are called in the end

enemy.cpp
TheoPannetier: superfluous call, I think?

food.h
```diff
@@ -47,6 +49,15 @@ int get_greenness(const food &f) noexcept;
 /// Get the redness (from the color) of the food
 int get_redness(const food &f) noexcept;
 
+
+
```
TheoPannetier, line 53:
```suggestion
```

game_view.cpp
```diff
@@ -279,10 +279,15 @@ void game_view::draw_player_coords() noexcept
...
-        str_player_coords += "Player " + p.get_ID() + " x = " + std::to_string(trunc(p.get_x()));
-        str_player_coords += "\nPlayer " + p.get_ID() + " y = " + std::to_string(trunc(p.get_y()));
+        str_player_coords += "Player " + p.get_ID() + " x = " + std::to_string(trunc(get_x(p)));
+        str_player_coords += "\nPlayer " + p.get_ID() + " y = " + std::to_string(trunc(get_y(p)));
         str_player_coords += "\n\n";
     }
+    food f = m_game.get_food()[0];
+    str_player_coords += "Food x = " + std::to_string(trunc(get_x(f)));
+    str_player_coords += "y = " + std::to_string(trunc(get_y(f)));
+    str_player_coords += "\n\n";
```
TheoPannetier, line 289: Do we want to keep this? I believe it may still be useful for dev for now but it is placed awkwardly.

TheoPannetier (approved):
Brilliant work, thanks for fixing the random position 👍

TheoPannetier:
I've reviewed the changes you guys made, @swom could you review the parts I changed?

game_view.cpp
swom: yeah maybe in another issue

swom (commented):
accepted changes suggested

swom:
fastest answer on this repo

TheoPannetier:
@swom I meant the commits I made earlier in this PR 😅 
Are you happy with my solutions for #411?

swom:
> @swom I meant the commits I made earlier in this PR 😅 Are you happy with my solutions for #411?

Ahhhhhgh, too much stuff on a single pull request, will go back and check, where is it?

TheoPannetier:
> Ahhhhhgh, too much stuff on a single pull request

Yeah, sorry I'll switch to another branch earlier next time

> where is it?

Kinda everywhere. Use the file changed feature of the PR 😉

swom:
Ah yeah ok, you did not implement templates for now. Yeah all seems ok. Will not check in detail and if things are not ok, I will take care.

TheoPannetier:
> 
> 
> There is a way to improve the test for #257, probably @Clemcarle can help. Basically just stitch a for loop around lines `1717` up to `1723` and repeat it a few thousand times, collect the `coordinate`s generated inside the loop in a `vector` and then check that those `coordinate`s' `x` and `y` have mean and variance corresponding to the mean and variance of an uniform distribution with the same ranges of `x` and `y`. @richelbilderbeek do you think there is another more elegant way to prove that they are coming from a distribution?

Just saw this. We actually discussed this with @richelbilderbeek in 20918be, for #250 .
The thing is the test should remain as simple as possible, and the issue doesn't specify that new positions should be sampled in a distribution, just that they should be different. So we stick to the concise and very simple test.

swom:
> > There is a way to improve the test for #257, probably @Clemcarle can help. Basically just stitch a for loop around lines `1717` up to `1723` and repeat it a few thousand times, collect the `coordinate`s generated inside the loop in a `vector` and then check that those `coordinate`s' `x` and `y` have mean and variance corresponding to the mean and variance of an uniform distribution with the same ranges of `x` and `y`. @richelbilderbeek do you think there is another more elegant way to prove that they are coming from a distribution?
> 
> Just saw this. We actually discussed this with @richelbilderbeek in [20918be](https://github.com/tresinformal/game/commit/20918bed3a666ade0baae8b514ba09ed9c1728b6), for #250 . The thing is the test should remain as simple as possible, and the issue doesn't specify that new positions should be sampled in a distribution, just that they should be different. So we stick to the concise and very simple test.

I actually think the original test went as I re-suggested, I remember this conversation. Good. All done then 👍

TheoPannetier:
> 
> 
> Ah yeah ok, you did not implement templates for now. Yeah all seems ok. Will not check in detail and if things are not ok, I will take care.

I did actually use the templated functions in most cases, except where a call to a member function was necessary 😇

Issue #353, opened 2021-07-29
Make sure the same keys are not drawn twice for different players


TheoPannetier:
For testing purposes, we use `get_random_game_options` to draw random keys for the different actions.

I have made sure that a key is not drawn twice for the same player (e.g., A corresponding to both turning left and right), but the same key may be drawn for different players (e.g. A makes player 1 turn left and player 2 go up). 

Make sure this doesn't happen!

```cpp
#ifdef FIX_ISSUE_353
  {
    // Random game options should not draw keys already used by another player
    const int rng_seed = 271; // by default this will assign A to both players
    const game_options rgo = get_random_game_options(rng_seed);
    key_action_map map_1 = rgo.get_kam_1();
    key_action_map map_2 = rgo.get_kam_2();
    bool a = map_1.has_key(sf::Keyboard::Key::A);
    bool b = map_2.has_key(sf::Keyboard::Key::A);
    // Key A is in KAM1 or KAM2 but not both
    assert(a != b);
  }
  #endif // FIX_ISSUE_353
```

TheoPannetier:
I believe this has been done in #431

PR #431, opened 2021-11-17
Solved issue #353

Clemcarle:
Issue #353 has been solved!

### DIFF
--- a/game_options.cpp
+++ b/game_options.cpp
@@ -47,6 +47,7 @@ void music_on(game_options& o) noexcept
 
 game_options get_random_game_options(const int& rng_seed) {
   std::srand(rng_seed);
+// Need to create all KAMs at once to ensure all keys are unique over all KAMs
   std::vector<key_action_map> random_kams = get_n_random_kams(2);
   game_options random_game_options = game_options(
         rng_seed,
--- a/game_options.cpp
+++ b/game_options.cpp
@@ -150,18 +150,23 @@ void test_game_options()
   }
   #endif // FIX_ISSUE_303
 
-   #define FIX_ISSUE_353
+  #define FIX_ISSUE_353
   #ifdef FIX_ISSUE_353
   {
     // Random game options should not draw keys already used by another player
-    const int rng_seed = 271; // by default this will assign A to both players
-    const game_options rgo = get_random_game_options(rng_seed);
-    key_action_map map_1 = rgo.get_kam_1();
-    key_action_map map_2 = rgo.get_kam_2();
-    bool a = map_1.has_key(sf::Keyboard::Key::A);
-    bool b = map_2.has_key(sf::Keyboard::Key::A);
-    // Key A is in KAM1 or KAM2 but not both
-    assert(a != b);
+    const int rng_seed = 271;
+    for (int seed = 0; seed < 1000; seed++)
+      {
+        const game_options rgo = get_random_game_options(rng_seed);
+        key_action_map map_1 = rgo.get_kam_1();
+        key_action_map map_2 = rgo.get_kam_2();
+        std::vector<sf::Keyboard::Key> keys_in_1;
+        for (auto& this_row : map_1.get_raw_map())
+          {
+            sf::Keyboard::Key key_in_1 = this_row.first;
+            assert(!map_2.has_key(key_in_1));
+          }
+      }
   }
   #endif // FIX_ISSUE_353
 
--- a/game_options.cpp
+++ b/game_options.cpp
@@ -147,7 +147,7 @@ void test_game_options()
   }
   #endif // FIX_ISSUE_303
 
-  // #define FIX_ISSUE_353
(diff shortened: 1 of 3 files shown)


Clemcarle:
So this is quite mysterious, because it runs succesfully on my and Oliver's computers (Windows), but fails on Richel's and on github (Linux). Anyone has any idea what is going on?

TheoPannetier:
This is odd indeed. I suspect this has to do with the RNG drawing different keys on different machines. Also the RNG itself might have changed since I wrote this test, so that the expected A is no longer drawn for either KAM. If so, I'll need to make a better test. I'll have a look sometime!

richelbilderbeek:
I have checked the code. It is not odd that different key maps have the same key; it is only checked if *a* keymap has different keys :-)

Clemcarle:
I am not sure about this, Richel: Making sure that different key maps can be generated will not have the same keys is the whole purpose of the `get_n_random_kams` function. Do you believe it is not doing that ?

richelbilderbeek:
@Clemcarle you are right and I was wrong: I see on Linux that the key is in neither :-)

Screenshot from 2021-11-19 11-18-54

TheoPannetier:
Ok, so my test is bad then: I assumed the same key would be drawn across all machines.
I'll come up with a better test 👍

richelbilderbeek:
It is a fun failed test :-)

TheoPannetier (approved):
Brilliant work, I like your solution of drawing both random KAMs at once 👍

TheoPannetier:
@richelbilderbeek what do you think of my new test for #353 ?
It's ugly because we run the test over 1000 semi-random iterations, but I don't see any other way to ensure that the test will work for any OS

richelbilderbeek:
Hi @TheoPannetier, I like the test! Doing it once, however, is good enough here :-). 

A hint about that is the number of tests, 1000, of which the number probably already felt arbitrary :-)

game_options.cpp
```diff
@@ -47,16 +47,19 @@ void music_on(game_options& o) noexcept
 
 game_options get_random_game_options(const int& rng_seed) {
   std::srand(rng_seed);
+  std::vector<key_action_map> random_kams = get_n_random_kams(2);
```
richelbilderbeek, line 50: I see why this is done. I suggest a comment such as:

```c++
// Need to create all KAMs at once to ensure all keys are unique over all KAMs
```

game_options.cpp
```diff
@@ -147,18 +150,23 @@ void test_game_options()
...
-    const int rng_seed = 271; // by default this will assign A to both players
-    const game_options rgo = get_random_game_options(rng_seed);
-    key_action_map map_1 = rgo.get_kam_1();
-    key_action_map map_2 = rgo.get_kam_2();
-    bool a = map_1.has_key(sf::Keyboard::Key::A);
-    bool b = map_2.has_key(sf::Keyboard::Key::A);
-    // Key A is in KAM1 or KAM2 but not both
-    assert(a != b);
+    const int rng_seed = 271;
+    for (int seed = 0; seed < 1000; seed++)
```
richelbilderbeek, line 158: I enjoy the thoroughness of 1000 iterations! Still, this costs the project in the long run, as everytime the test suit is run, there are 999 superfluous iterations. Also, the feeling you got when picking 1000, i.e. 'Why 1000? Why no 100? Or 10? Or 10000?' is an indication that this was an arbitrary decision. 1 time is good enough here :-)

key_action_map.cpp
```diff
@@ -106,9 +106,9 @@ key_action_map get_random_kam()
             new_key
             );
       if (key_match == v_random_keys.end())
-        {
+      {
```
richelbilderbeek, line 109: Yay, the boy scout rule!

key_action_map.cpp
```diff
@@ -120,6 +120,43 @@ key_action_map get_random_kam()
       );
 }
 
+std::vector<key_action_map> get_n_random_kams(int n)
+{
+  std::vector<key_action_map> vector_kam;
```
richelbilderbeek, line 125: To be examplary, add `assert(n >= 0);`, else the for loop will do approx 2^64 too much iterations :-)

richelbilderbeek (changes requested):
I enjoy seeing this work. I have some minor comments, but I feel they are important enough to pick 'Request Changes', also because I think you want to do excellent. Feel free to debate and ask me for the next review! Keep up the good job!

key_action_map.cpp
TheoPannetier:
```suggestion
  assert(n >= 0);
  std::vector<key_action_map> vector_kam;
```

game_options.cpp
TheoPannetier: Done!

richelbilderbeek (approved):
Happy to see you could agree with my proposed changes!

Issue #406, opened 2021-10-21
The positions of all shelters can be accessed easily


TheoPannetier:
Depends on:
- [x] #405 

For #315, possibly other utilities we would need to be able to collect the positions of all shelters in a `game`.
Make a function `get_all_shelter_positions()` that takes care of this.

Protip: use `get_nth_shelter_position()` from #405 to do this 😉 

```cpp
//#define FIX_ISSUE_406
  #ifdef FIX_ISSUE_406
  {
    // the position of all shelters can be obtained
    
    double wall_short_side = 1600;
    int num_players = 0;
    int n_ticks = 0;
    int n_shelters = 5;
    
    game g(wall_short_side, num_players, n_ticks, n_shelters);
    
    std::vector<coordinate> expected_shelter_positions;
    for (int n = 0; n < n_shelters; ++n)
      {
        coordinate nth_shelter_position = get_nth_shelter_position(g, n);
        expected_shelter_positions.push_back(nth_shelter_position);
      }
    
    std::vector<coordinate> shelter_positions = get_all_shelter_positions(g);
    
    assert(shelter_positions.size() == expected_shelter_positions.size());
    for (int n = 0; n < n_shelters; ++n)
      {
        assert(shelter_positions[n] == expected_shelter_positions[n]);
      }
  }
```

TheoPannetier:
Done!

Issue #405, opened 2021-10-21
A shelter's position can be accessed easily


TheoPannetier:
The position of a shelter can be accessed with `g.get_shelters()[n].get_position()`, but this is cumbersome to read.
Wrap this inside a shorthand function called `get_nth_shelter_position()`

```cpp
// #define FIX_ISSUE_405
  #ifdef FIX_ISSUE_405
  {
    // nth shelter position can be obtained
    game g;
    int n = 0;
    shelter first_shelter = g.get_shelters()[n];
    coordinate expected_c = first_shelter.get_position();
    
    coordinate c = get_nth_shelter_position(g, n);
    
    assert(c == expected_c);
  }
  #endif
```

TheoPannetier:
Done!

PR #436, opened 2021-11-17
Christiaan

TheoPannetier:
- #405 
- #406

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -771,7 +771,7 @@ std::vector<coordinate> get_all_shelter_positions(const game& g)
 {
   std::vector<shelter> all_shelters = g.get_shelters();
   std::vector<coordinate> all_shelter_positions;
-  for (;;)
+  for (unsigned int i = 0;all_shelters.size() != i;)
     {
 
     }
--- a/game.cpp
+++ b/game.cpp
@@ -746,6 +746,7 @@ int get_nth_food_regeneration_time(const game &g, const int &n)
   return g.get_food()[n].get_regeneration_time();
 }
 
+
 bool is_nth_food_eaten(const game& g, const int &n)
 {
   return g.get_food()[n].is_eaten();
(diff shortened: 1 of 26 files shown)


game.cpp
```diff
@@ -11,7 +11,7 @@
 #include <iostream>
 #include <random>
 
-game::game(double wall_short_side,
+game::game(environment the_environment,
```
richelbilderbeek, line 14: Nice, a new feature! What would be even better, though, is to use `const& environment`, see [the C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in).

game.cpp
```diff
@@ -761,6 +762,23 @@ void place_nth_food_randomly(game &g, const int &n)
   g.get_food()[n].place_randomly(g.get_rng(), {get_min_x(g), get_min_y(g)}, {get_max_x(g), get_max_y(g)});
 }
 
+coordinate get_nth_shelter_position(const game &g, const int &n)
+{
+  return g.get_shelters()[n].get_position();
```
richelbilderbeek, line 767: Nice first setup. To be examplary, don't forget that `i` needs to be positive, for `g.get_shelters()[n]` to make sense. Hence, adding `assert(n >= 0);` give a nice error message during debug.

game.cpp
```diff
@@ -761,6 +762,23 @@ void place_nth_food_randomly(game &g, const int &n)
...
+}
+
+/// Get all shelter positions
+std::vector<coordinate> get_all_shelter_positions(const game& g)
```
richelbilderbeek, line 771: Nice! One day, you'll learn algorithms and replace that 'dumb' for-loop by such an algorith, For now: well done!

richelbilderbeek (changes requested):
Well done! I have only one important comment about adding an `assert`, after that: hit me again and I will happily accept your work!

game.cpp
richelbilderbeek: Yay!

richelbilderbeek (approved):
Well done, thanks!

Issue #376, opened 2021-09-22
class 'read_only' returns an unnecessary copy


richelbilderbeek:
Class `read_only` has two member functions that (are too similar and) return a copy (`T`) instead of a const-reference (`const T&`):

```c++
class read_only 
{
  // ...
  T value() const // ...

   // ...

  T get() const noexcept // ...
};
```

Adapt the code to return a `const T&` instead. This should be very easy.

richelbilderbeek:
This is fixed!

PR #438, opened 2021-11-24
Fixed issue #376

Tingvall:


TheoPannetier (changes requested):
Nice start, but I think #376 is not solved yet - apologies if I'm mistaken! I believe one of the two functions could also be ditched, as suggested in #376.

Tingvall:
Me and Richel got it to work on mac, that is the new stuff. Ignore in review.

richelbilderbeek:
@Tingvall: I do agree that the `.pro` files can be ignored in the review, but the `#ifdef USE_MAC`s should be removed before merging to `develop` I think. Who wrote that in the first place :smirk:?

kausmees:
We removed one of the functions, and renamed the other one get_value().
Also removed the preprocessor #ifdef USE_MAC directives from all files, and it worked when we tested on a mac!

TheoPannetier (approved):
Excellent! #376 is clearly addressed here, no further comments.
Huge thanks for getting it to work for Mac users 🍎

Issue #432, opened 2021-11-17
Create an `about.cpp` and `about.h` that will contain the logic of the about screen.


swom:
Add this files to the game.pri project.

richelbilderbeek:
I've done so; I created an `about` class, with a list of all Collaborators. I think there is plenty of room for improvement, yet I suggest/act to make these into new Issues.

Done!

PR #445, opened 2021-12-08
Create an about class, progress #432

richelbilderbeek:
This is the first step in getting a simple `about` class. Except for that, I cleaned up the `.pri` file by sorting it alphabetically :-)

### DIFF
--- a/about.cpp
+++ b/about.cpp
@@ -1,0 +1,41 @@
+#include "about.h"
+
+about::about()
+{
+
+}
+
+std::vector<std::string> about::get_contributors() const noexcept
+{
+  // Ordered by list at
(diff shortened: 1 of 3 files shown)


about.cpp
```diff
@@ -0,0 +1,41 @@
...
+    "Xiaoyan Long",
+    "Chris Hesper",
+    "Manuel",
+    "Ines Daras",
+    "Johanna Tingvall-Gustafsson",
+    "Joshua Lambert",
+    "Kirsten Ferwerda",
+    "Tim Temizyuerek",
+    "Jan Derk Kotlarski",
+    "Timo",
```
swom, line 36:
```suggestion
    "Timo van Eldijk",
```

swom (approved):

PR #449, opened 2021-12-15
Fixing #441

### LINKED ISSUE #441: A player can have its color set
Write a test in player.cpp in which a player's color is set. Add a member function, `player::set_color` that does so.

This is needed for #440 

```cpp
//  #define FIX_ISSUE_441
  #ifdef FIX_ISSUE_441
    {
      // #441 A player's color can be set
      player p;
      // Turn the player green
      const color green = create_green_color();
      p.set_color(green);
      assert(p.get_color() == green);
      // Turn the player red
      const color red = create_red_color();
      p.set_color(red);
      assert(p.get_color() == red);
    }
  #endif
```

swom:
Fixed #441


richelbilderbeek (approved):
Excellent! There is one minor comment on adding `noexcept`, but this is too minor to request changes. Well done!

Issue #458, opened 2022-01-11
The `color` of any food item can be accessed easily


TheoPannetier:
To help making tests for implementing changes in the color of food items, it would be great to have a free function that helps accessing the color of any given food item in the game. Make such a function come to life 🌈🍪 
```cpp
//#define FIX_ISSUE_458
#ifdef FIX_ISSUE_458
  {
    // (458) The color of any food item can be accessed easily
    game g;
    color color_food = get_nth_food_color(g, 0);
    // Default color for the food item is black
    color c;
    assert(color_food == c);
  }
#endif
```

**Hint:** if you're stuck, look for existing functions that do similar stuff, e.g. `get_nth_food_x()`

Issue #457, opened 2022-01-11
The `color` of any player can be accessed easily


TheoPannetier:
To prepare the ground for implementing  changes in the color of players, it would be great to have a free function that helps accessing the color of any given player. Make such a function come to life 🌈 

```cpp
#define FIX_ISSUE_457
#ifdef FIX_ISSUE_457
  {
    // (457) The color of any player can be accessed easily
    game g;
    color color_player_one = get_nth_player_color(g, 0);
    color color_player_two = get_nth_player_color(g, 1);
    color color_player_three = get_nth_player_color(g, 2);
    // Default colors for players: R, G, B
    assert(color_player_one == create_red_color()); 
    assert(color_player_two == create_green_color());    
    assert(color_player_three == create_blue_color());        
  }
#endif
```

**Hint:** if you're stuck, look for existing functions that do similar stuff, e.g. `get_nth_player_size()`

swom:
@ollyturner @Yangshen0325 
Hello guys, I have decided to reassign you to this issue as you had completed it in your previous pull request but in a way that after some consideration I did not consider satisfying. Please try again to solve this issue (merge `develop` on your branch first) _without_ commenting out parts of the test.
Best of luck!

PR #459, opened 2022-01-11
tests for #457, #458

TheoPannetier:
**Tests for:**
- #457 
- #458

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -1917,18 +1917,15 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
-//#define FIX_ISSUE_457
-#ifdef FIX_ISSUE_457
+//#define FIX_ISSUE_458
+#ifdef FIX_ISSUE_458
   {
-    // (457) The color of any player can be accessed easily
+    // (458) The color of any food item can be accessed easily
     game g;
-    color color_player_one = get_nth_player_color(g, 0);
-    color color_player_two = get_nth_player_color(g, 1);
-    color color_player_three = get_nth_player_color(g, 2);
-    // Default colors for players: R, G, B
-    assert(color_player_one == create_red_color());
-    assert(color_player_two == create_green_color());
-    assert(color_player_three == create_blue_color());
+    color color_food = get_nth_food_color(g, 0);
+    // Default color for the food item is black
+    color c;
+    assert(color_food == c);
   }
 #endif
 
--- a/game.cpp
+++ b/game.cpp
@@ -1906,10 +1906,10 @@ void test_game() //!OCLINT tests may be many
 #ifdef FIX_ISSUE_457
   {
     // (457) The color of any player can be accessed easily
-    game g;
-    color color_player_one = get_nth_player_color(g, 0);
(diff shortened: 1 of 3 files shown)


game.cpp
```diff
@@ -1902,6 +1902,33 @@ void test_game() //!OCLINT tests may be many
...
+
+//#define FIX_ISSUE_458
+#ifdef FIX_ISSUE_458
+  {
+    // (458) The color of any food item can be accessed easily
+    game g;
+    color color_food = get_nth_food_color(g, 0);
+    // Default color for the food item is black
+    color c;
+    assert(color_food == c);
```
richelbilderbeek, line 1928: This is scary, as the comment is needed to explain the code, where these may very likely change in the future.

```c++
assert(color_food == color::black);
assert(color_food == create_black());
assert(color_food == create_black_color());
```

game.cpp
```diff
@@ -1902,6 +1902,33 @@ void test_game() //!OCLINT tests may be many
...
+    assert(color_player_two == create_green_color());
+    assert(color_player_three == create_blue_color());
+  }
+#endif
+
+//#define FIX_ISSUE_458
+#ifdef FIX_ISSUE_458
+  {
+    // (458) The color of any food item can be accessed easily
+    game g;
```
richelbilderbeek, line 1924: As we only read from the `game`, it can (and hence should) be `const game g;`

game.cpp
```diff
@@ -1902,6 +1902,33 @@ void test_game() //!OCLINT tests may be many
   }
   #endif // FIX_ISSUE_241
 
+//#define FIX_ISSUE_457
+#ifdef FIX_ISSUE_457
+  {
+    // (457) The color of any player can be accessed easily
+    game g;
```
richelbilderbeek, line 1909: As we only read from the `game`, it can (and hence should) be `const game g;`

game.cpp
```diff
@@ -1902,6 +1902,33 @@ void test_game() //!OCLINT tests may be many
...
+    color color_player_one = get_nth_player_color(g, 0);
+    color color_player_two = get_nth_player_color(g, 1);
```
richelbilderbeek, line 1911: All these colors can (and hence should) be `const`

richelbilderbeek (changes requested):
Hi @TheoPannetier,

Great work! 

I have been a bit more strict as I think you will like that and hence request to be more const-correct, especially [Con.1: By default, make objects immutable](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable).

So, get those `const`s in and I'd be happy to accept!

game.cpp
TheoPannetier: The test now expects the food color to be the default color (of the class). No need for a possibly lying comment anymore!

TheoPannetier:
I've sneaked a new test in, for #460

richelbilderbeek (approved):
Increased const-correctness, I like that! Note that `is_first_color_winner` could also be const-corrected, now it is time to get this on develop :-)

richelbilderbeek:
Well done @TheoPannetier !

Issue #464, opened 2022-01-18
A player's state can be accessed easily


TheoPannetier:
It would be handy to be able to access a player's state in one line, e.g. for the test in #463 .
Write a free function that returns member `m_player_state` for the `n`th player in a `game`.

**Hint**: if you're struggling or don't know where to start, have a look at similar, existing functions, e.g. `get_nth_player_size()` 😉 

```cpp
//#define FIX_ISSUE_464
#ifdef FIX_ISSUE_464
  {
    // (464) A player's state can be accessed easily
    const game g;
    assert(get_nth_player_state(g, 0) ==  g.get_players()[0].get_state());
  }
#endif
```

PR #465, opened 2022-01-18
Theo

TheoPannetier:
Tests for

- #463 
- #464

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -1139,7 +1139,7 @@ void test_game() //!OCLINT tests may be many
   // Players lose invulnerability after a short time
   {
     game g;
-    int duration_invulnerability = 3;
+    const int duration_invulnerability = 3;
     // change the nb above to a value you deem reasonable
     // after fixing the test delete these two comment lines
     become_invulnerable(g.get_player(0));
--- a/game.cpp
+++ b/game.cpp
@@ -1947,6 +1947,16 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
+//#define FIX_ISSUE_464
+#ifdef FIX_ISSUE_464
+  {
+    // (464) A player's state can be accessed easily
+    const game g;
+    const player_state default_state = player_state::active;
+    assert(get_nth_player_state(g, 0) == default_state);
+  }
+#endif
+
 #endif // no tests in release
 }
 
--- a/game.cpp
+++ b/game.cpp
@@ -1134,6 +1134,25 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
+//#define FIX_ISSUE_463
+#ifdef FIX_ISSUE_463
+  // Players lose invulnerability after a short time
+  {
+    game g;
+    int duration_invulnerability = 3;
+    // change the nb above to a value you deem reasonable
+    // after fixing the test delete these two comment lines
+    become_invulnerable(g.get_player(0));
+    for (int t = 0; t < duration_invulnerability; t++)
+      {
+        assert(is_invulnerable(g.get_player(0)));
+        g.tick();
+      }
+    assert(!is_invulnerable(g.get_player(0)));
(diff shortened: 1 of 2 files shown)


game.cpp
```diff
@@ -1134,6 +1134,25 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
+//#define FIX_ISSUE_463
+#ifdef FIX_ISSUE_463
+  // Players lose invulnerability after a short time
+  {
+    game g;
+    const int duration_invulnerability = 3;
```
richelbilderbeek, line 1142: This constant should in the future be part of `game`/`game_options` class. This is a good first step, with helpful comments.

game.cpp
```diff
@@ -1928,6 +1947,16 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
+//#define FIX_ISSUE_464
+#ifdef FIX_ISSUE_464
+  {
+    // (464) A player's state can be accessed easily
+    const game g;
+    const player_state default_state = player_state::active;
+    assert(get_nth_player_state(g, 0) == default_state);
```
richelbilderbeek, line 1956: I enjoy the attempt and I think we both see how brittle this test is: it depends on the game's default state.

I suggest to compare the results of this logistic (as in, getting data from A to B) function with the longer way that is also possible, e.g. something like:

```c++
assert(get_nth_player_state(g, 0) == g.get_players()[0].get_state()); 
```

In that way, the logistic function is tested to be that actual shortcut.

It may feel like code duplication if the test does the same as the logistic function, but it isn't, as functions change. This test will guarantee the logisistic function doing the right thing.

game.cpp
```diff
@@ -1134,6 +1134,25 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
+//#define FIX_ISSUE_463
+#ifdef FIX_ISSUE_463
```
richelbilderbeek, line 1138: After the tests are done, remove the preprocessor statements (i.e. `#define`, `#ifdef`, `#endif`). They are there to put tests in the code that fail (so our Issue can point to them). When the tests passes, these can be removed.

I do usually make it a newbie Issue to remove these, so maybe that is your strategy as well :-)

game.cpp
richelbilderbeek: Seems like a nice thing to create an Issue for!

game.cpp
richelbilderbeek: Another way to remove the brittleness:

```c++
const player_state default_state = get_default_player_state();
```

richelbilderbeek: I suggest your pick none, one or both of my suggested solutions and create an Issue for it :+1:

richelbilderbeek (changes requested):
Good stuff! 

I do not necessarily request changes; it is good enough to be merged. Instead, here is the opportunity to improve the code from my feedback, or create Issues. You go and pick!

Next review request, I will simply accept :-)

game.cpp
TheoPannetier: Agreed! Implementing this member variable is actually an implicit step of the issue. But I guess issues shouldn't have implicit steps!
I'd like to push the tests first in case someone wants to tackle it today, but I'll rework it to implement this if nobody takes it

game.cpp
```diff
@@ -1134,6 +1134,23 @@ void test_game() //!OCLINT tests may be many
...
+  // Players lose invulnerability after a short time
+  {
+    game g;
+    player& p = g.get_player(0);
```
TheoPannetier, line 1142: @richelbilderbeek is using a ref here a good idea?
It saves a lot of retyping `g.get_player(0)` below, but I'm unsure if this helps or hinder readability.

richelbilderbeek: @TheoPannetier: If it increases readablity, yes, you can opt to do so :-)

Issue #474, opened 2022-01-26
Make code independent of SFML version


kausmees:
With SFML version 2.4, the function sf::Text::SetColor() was deprecated and replaced by other ones, including setFillColor() which is currently used in the code.

(see https://www.sfml-dev.org/documentation/2.4.0/classsf_1_1Text.php)

This causes a build error when using a SFML version < 2.4

Preprocessor directives should be used to handle this so that the build works for SFML versions younger than 2.4.

TheoPannetier:
Done, thanks @kausmees !

PR #475, opened 2022-01-26
Fix issue #474

kausmees:
Added preprocessor directives to use different functions dependent on the SFML version.

### DIFF
--- a/menu_view.cpp
+++ b/menu_view.cpp
@@ -89,12 +89,12 @@ void menu_view::draw_buttons() noexcept
     button_text.setPosition(static_cast<float>(button_position.get_x()),
                      static_cast<float>(button_position.get_y()));
     button_text.setCharacterSize(30);
-#if SFML_VERSION_MAJOR >= 2 and SFML_VERSION_MINOR >= 4
-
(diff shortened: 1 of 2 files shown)


kausmees:
Opps, I realized this code won't work for SFML major versions > 2 (then there should be no constraint on the minor version), so that should probably be fixed before accepting the pull request.

richelbilderbeek:
@kausmees the tests pass and we have the mantra 'If all tests pass, we are -by definition- happy'. Sure, if team members discover their build fails, we'll fix it :-)

TheoPannetier:
> Oops, I realized this code won't work for SFML major versions > 2 (then there should be no constraint on the minor version), so that should probably be fixed before accepting the pull request.

I pushed a change solving this with an elif 😉

TheoPannetier (approved):
Brilliant work, thanks!
The only downside I see is that every time we'll want to colour something we'll have to add these preprocessor statements, and they could take a lot of lines if we have many things to colour... One could make a custom colouring function so these are only evaluated once in the entire code, but I think it's not an issue rn.

Issue #466, opened 2022-01-18
[Housekeeping] Move free functions to separate file and header


TheoPannetier:
☁️ :broom:
So far, most (all?) free getter functions like `get_nth_food_color()`, `get_nth_player_size()` etc. are located in `game.cpp`.
While it makes sense to include them there because they all take a `game` as their first argument, they do not belong to the actual class `game` and take up a significant amount of space, along with their tests. It would be great to move them to a separate file (e.g. `free_functions.cpp` or `free_getters.cpp` ?) that also include their tests, so `game.cpp` could be a bit more readable.

richelbilderbeek:
I like this idea! 

I would need to look up the recommended practices here. I feel one big file called `free_functions.h` would only move the problem, as other classes also have free functions. A `game_functions.h` appeals more to me. Again, need to look up in The C++ Core Guidelines!

As you spotted the free functions that take a `game` are there, but they need not be there when we will discuss forward declarations today.

TheoPannetier:
Ok, looking forward to hear about forward declarations 😀

PR #476, opened 2022-02-01
Moved free functions to `game_functions`

TheoPannetier:
- [ ] #466

\+ changed some functions names to be more specific

### DIFF
--- a/game.h
+++ b/game.h
@@ -38,14 +38,7 @@ public:
   void do_actions() noexcept;
 
   ///Executes for wall collision for all players
-  void do_wall_collisions();
-
-  ///Get the index of the winner of the players with the given indices
-  int get_winning_player_index(const game& g, const int i1, const int i2);
-
-
-  ///Get the index of the loser of the players with the given indices
-  int get_losing_player_index(const game& g, const int i1, const int i2);
+  void resolve_wall_collisions();
 
   ///returns the collision vector
   const std::vector<int>& get_collision_vec(){return m_v_collisions_ind;}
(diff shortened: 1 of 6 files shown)


game.h
```diff
@@ -127,7 +117,7 @@ class game
   int get_dist_x_pls() const noexcept { return m_dist_x_pls; }
 
   ///Manages collisons with walls
-  player wall_collision(player p);
+  player resolve_wall_collision(player p);
```
richelbilderbeek, line 120: From the function prototype, I find it hard to guess what it does: it takes a player, modifies it and puts it back in game I guess? This is one case where `void resolve_wall_collision(player& p);` would be better. Also, why is it not a const member function (as I guess the arena does not change?).

Still, this has nothing to do with the PR, so I will accept :-)

game_functions.cpp
```diff
@@ -0,0 +1,226 @@
+#include <game_functions.h>
+#include <coordinate.h>
+#include <game.h>
+
+// Game functions are free functions that return some info about the game but do not change it
+// Formally, if a function
+// 1. Isn't inside a class
+// 2. Takes argument game
+// 3. Is passed a const ref to  game (const game&)
+// Then it should be defined here. Otherwise not.
```
richelbilderbeek, line 10: I enjoy making these rules explicit!

game_functions.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <coordinate.h>
+#include <game.h>
+
+// Game functions are free functions that return some info about the game but do not change it
+// Formally, if a function
+// 1. Isn't inside a class
+// 2. Takes argument game
+// 3. Is passed a const ref to  game (const game&)
+// Then it should be declared here. Otherwise not.
```
richelbilderbeek, line 12: Yay, explicit rules!

richelbilderbeek (approved):
I am happy with this housekeeping. There were problems in the original code, that I ignore for now, as that is not the goal of this PR. Regardless of that, I see some love added to the original code. Thanks!

Issue #484, opened 2022-02-06
One can detect if a position is inside or outside a menu button


TheoPannetier:
Depends on:
- [x] #482 

```cpp
#ifdef FIX_ISSUE_484
  {
    // (484) void test_menu_view()
    const menu m;
    const menu_button mb = m.get_button("about");
    const float mb_width = mb.get_body().x;
    const float mb_height = mb.get_body().y;
    const coordinate c_inside(
          mb.get_x() + mb_width / 2.0,
          mb.get_y() + mb_height / 2.0
          );
    const coordinate c_outside1(
          mb.get_x() + mb_width / 2.0 + 1.0,
          mb.get_y() + mb_height / 2.0
          );
    const coordinate c_outside2(
          mb.get_x() + mb_width / 2.0,
          mb.get_y() + mb_height / 2.0 + 1.0
          );
    assert(is_inside_button(c_inside, "about"));
    assert(!is_inside_button(c_outside1, "about"));
    assert(!is_inside_button(c_outside2, "about"));
  }
#endif // FIX_ISSUE_484
```

TheoPannetier:
Done!

Issue #483, opened 2022-02-06
Calling a button that doesn't exist returns an error


TheoPannetier:
Depends on
- [x] #482 

Look for the solution for #340 in the code to see how you can throw an error ⚠️ 

```cpp
#ifdef FIX_ISSUE_483
  {
    // (483) Calling a button that doesn't exist causes an error
    menu m;
    std::string wrong_label = "whatever, doesn't exist";
    try {
      menu_button mb = m.get_button(wrong_label); // throws exception
    }
    catch ( const std::exception& e ) {
      assert(std::string(e.what()) == std::string("No button in menu has this label."));
    }
  }
#endif
```

TheoPannetier:
Done!

Issue #482, opened 2022-02-06
A `menu_button` can be called from its label


TheoPannetier:
```cpp
  {
    // (482) Menu button can be called from its label
    menu m;
    std::string label = "about";
    menu_button mb_about = m.get_button(label);
    assert(mb_about.get_label() == label);
  }
```

TheoPannetier:
Done !

Issue #481, opened 2022-02-06
`menu_button` has a coordinate member


TheoPannetier:
So far `menu_button` has double members `x` and `y`, it would be nice to replace them with a `coordinate` member to be consistent with the rest of the game

PR #485, opened 2022-02-06
Menu buttons

TheoPannetier:
Aiming towards #450 
So far, clicking anywhere in the menu will close the game. In the future, I want to close the game if clicking happens inside the button "quit". To this end I've made progress towards getting a function `is_inside_button()` that would detect whether a position (e.g. the mouse's) is inside a button

Tested + fixed:

- [x] #482 
- [x] #483 
- [x] #481 
- [x] #484 

Fixed:

- [x] #450

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -28,10 +28,21 @@ std::vector<menu_button> &menu::get_buttons() noexcept { return m_v_buttons; }
 
 menu_button &menu::get_button(int index)
 {
-
   assert(index >= 0 && index < static_cast<int>(m_v_buttons.size()));
   return m_v_buttons[static_cast<unsigned int>(index)];
+}
 
+menu_button &menu::get_button(std::string label)
+{
+  for (auto& mb : m_v_buttons)
+   {
+    if (mb.get_label() == label)
+      {
+        return mb;
+      }
+   }
+  // If no match return an error
+  throw std::logic_error("No button in menu has this label.");
 }
 
 void menu::put_buttons_tidy() noexcept
@@ -73,7 +84,6 @@ void test_menu()
     menu_button mb_about = m.get_button(label);
     assert(mb_about.get_label() == label);
   }
-#ifdef FIX_ISSUE_483
   {
     // (483) Calling a button that doesn't exist causes an error
     menu m;
@@ -85,7 +95,6 @@ void test_menu()
       assert(std::string(e.what()) == std::string("No button in menu has this label."));
     }
   }
-#endif
   // buttons are evenly distributed
   // along the height of the screen
   {
--- a/menu.cpp
+++ b/menu.cpp
@@ -35,12 +35,12 @@ menu_button &menu::get_button(int index)
 menu_button &menu::get_button(const std::string& label)
 {
   for (auto& mb : m_v_buttons)
-   {
-    if (mb.get_label() == label)
-      {
-        return mb;
-      }
-   }
+    {
+      if (mb.get_label() == label)
+        {
+          return mb;
+        }
+    }
   // If no match return an error
   throw std::logic_error("No button in menu has this label.");
 }
--- a/menu.cpp
+++ b/menu.cpp
@@ -124,7 +124,7 @@ void test_menu()
   {
     // (484) void test_menu_view()
     const menu m;
-    const menu_button mb = m.get_button("about");
+    const menu_button mb = m.get_button(0);
     const float mb_width = mb.get_body().x;
     const float mb_height = mb.get_body().y;
     const coordinate c_inside(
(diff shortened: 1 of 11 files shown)


menu.cpp
```diff
@@ -64,7 +88,25 @@ void test_menu()
...
+    // (482) Menu button can be called from its label
+    menu m;
+    const std::string label = "about";
+    const menu_button mb_about = m.get_button(label);
+    assert(mb_about.get_label() == label);
+  }
+  {
+    // (483) Calling a button that doesn't exist causes an error
+    menu m;
+    const std::string wrong_label = "whatever, doesn't exist";
```
richelbilderbeek, line 103: I like this text :-)

menu.cpp
```diff
@@ -28,10 +28,21 @@ std::vector<menu_button> &menu::get_buttons() noexcept { return m_v_buttons; }
...
 menu_button &menu::get_button(int index)
 {
-
   assert(index >= 0 && index < static_cast<int>(m_v_buttons.size()));
   return m_v_buttons[static_cast<unsigned int>(index)];
+}
 
+menu_button &menu::get_button(const std::string& label)
+{
+  for (auto& mb : m_v_buttons)
```
richelbilderbeek, line 37: The indentation is a bit different than the usual, this is the most-user indentation structure:

```c++
for (..)
{
  if (...) 
  {

  }
}
```

menu.cpp
```diff
@@ -64,7 +88,25 @@ void test_menu()
   // could I use try and catch framework for exceptions in this case?
   {
     menu v;
-    assert(v.get_button(0).get_name() == "action");
+    assert(v.get_button(0).get_label() == "action");
+  }
+  {
+    // (482) Menu button can be called from its label
+    menu m;
```
richelbilderbeek, line 95: Ideally, it should be `const menu v`, as you only read from the menu. I see why: `get_button` is non-const, which is reasonable enough. There is a solution, however, which is *adding* a const version:

```c++
class menu_view
{
  // ...
  button& get_button(const int index); // Currently there, signature and name probably incorrect

  const button& get_button(const int index) const; // To be added
}
```

Because I think you'll enjoy that new piece of knowledge, I will not approve this Pull Request, so you can improve even more on your const-correctness :-)

menu.h
```diff
@@ -27,6 +27,9 @@ class menu
   /// Get one button at index i
   menu_button &get_button(int index);
 
+  /// Get one button at index i
+  menu_button &get_button(const std::string& label);
```
richelbilderbeek, line 31: As stated above: add the `const` member function of the same name here :-)

menu_button.h
```diff
@@ -3,37 +3,40 @@
...
+#include "coordinate.h"
 
 #include <cassert>
 #include <string>
 
 class menu_button
 {
 public:
-  menu_button(std::string name,
+  menu_button(std::string label,
```
richelbilderbeek, line 14: Why not `const std::string& label`? You may even be correct (but I keep the why a secret for now :-) ), I mostly check up on the reasoning here :-)

menu_view.cpp
```diff
@@ -1,6 +1,7 @@
 #include "menu_view.h"
 #include "coordinate.h"
 
+#ifndef LOGIC_ONLY // that is, compiled on GitHub Actions
```
richelbilderbeek, line 4: I could not have predicted this `ifndef` would be a good idea, but I assume you ran into build problems and found out that this was a solution. Well done!

menu_view.cpp
```diff
@@ -59,24 +83,23 @@ void menu_view::draw_buttons() noexcept
...
-    coordinate button_position(m_menu.get_button(i).get_x(),
-                               m_menu.get_button(i).get_y());
+    const coordinate button_position = this_button.get_position();
 
     // Create the button sprite
-    sf::RectangleShape rect(m_menu.get_button(i).get_body());
+    sf::RectangleShape rect(this_button.get_body());
     rect.setFillColor(button_color);
-    rect.setOrigin(rect.getSize().x / 2.0, rect.getSize().y / 2.0);
+    rect.setOrigin(rect.getSize().x / 2.0f, rect.getSize().y / 2.0f);
```
richelbilderbeek, line 102: I see the use of adding the `f` here, a good discovery!

richelbilderbeek (changes requested):
This is a great PR! I am happy to see the const-correctness is moving to a higher level and see some new C++ discoveries put into code! 

I will request changes to encourage you to learn something I think you'd enjoy to know! Even if you'd skip that opportunity, I'd accept on the next review request.

menu.cpp
TheoPannetier: On it!
But why does one need to add `const` both before and after the declaration?

TheoPannetier: Ah, helaas pindakaas I can't get the function to return a `const` output. I think one would need to make `m_v_buttons` a const variable as well, but that's not what a member variable should be.
I'm very confused about how the compiler deals with const, I'll just ignore this and move on for now.

menu.h
TheoPannetier: Couldn't get it to work :/

menu_button.h
TheoPannetier: It's just a variable used to initialise the value of the corresponding member function. I guess if we made this `const` the member variable would inherit from this and we wouldn't be able to change it anymore? Which is not desirable

menu_view.cpp
TheoPannetier: No, this is let over from an attempt at adding a test block to `menu_view`. I have no idea what this is actually for, I'll just rm it

richelbilderbeek (approved):
Thanks!

PR #491, opened 2022-02-16
Fix the issue that game_resource.qrc is included twice

EvoLandEco:


game_helpers.h
```diff
@@ -0,0 +1,4 @@
+#ifndef GAME_HELPERS_H
+#define GAME_HELPERS_H
+
+#endif // GAME_HELPERS_H
```
swom, line 4: Mmm this seems like an empty file, maybe you can delete it. But we are always in need of game helpers.

game_view.pri
```diff
@@ -9,6 +9,3 @@ SOURCES += \
     $$PWD/game_view.cpp \
     $$PWD/menu_view.cpp \
     $$PWD/options_view.cpp \
-
-RESOURCES += \
-    game_resources.qrc
```
swom, line 14: seems legit! Well done! 👍

game_helpers.h
richelbilderbeek: Yeah, this code was already there, Tianjian has not added it, hence I consider this resolved.

PR #503, opened 2022-03-09
WIP issue #457

ollyturner:
We tried to fix Issue #341, but it seemed to run. Please double check to see if we need to add/change anything.

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -558,6 +558,13 @@ void put_player_near_food(player &p, const food &f, const double distance)
   p.place_to_position(new_position);
 }
 
+color get_nth_player_color(const game& game, const int player_id)
+{
+    player p = game.get_player(player_id);
+    color c = create_red_color();
+    return c;
+}
+
 void test_game() //!OCLINT tests may be many
 {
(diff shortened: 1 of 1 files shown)


richelbilderbeek:
Dear @ollyturner and @Yangshen0325, thanks for the code, I am sure it is good progress. However, that merge conflict must be fixed before anyone will properly review it. Please do so :-)

game.cpp
```diff
@@ -560,6 +560,13 @@ void put_player_near_food(player &p, const food &f, const double distance)
   p.place_to_position(new_position);
 }
 
+color get_nth_player_color(const game& game, const int player_id)
+{
+    player p = game.get_player(player_id);
+    color c = create_red_color();
+    return c;
```
swom, line 567: Indeed this makes the test pass, but not exactly the way we want it to pass. In this case no matter the player the color returned will always be red.

swom: I will accept this and make an issue.

swom (approved):
You sneaked your way through the test, but the war is not over.

Issue #456, opened 2022-01-10
Truncate the displayed coordinates


TheoPannetier:
![](https://raw.githubusercontent.com/tresinformal/game/master/pics/20220110.png)

The numbers showing object coordinates in the top-right corner are useful, but look a bit awkward. There are a bunch of decimals printed there, yet if you run the game you'll see that their value is always zero.
So these numbers don't carry any useful information for us, and impede reading. 
Find a way to either truncate the numbers (maybe by converting them to integers before drawing?), or make the string only display the integer part of the coordinate.

EvoLandEco:
trunc(get_x(p)) doesn't work, I guess it is because trunc() also returns a double? Explicit type conversion works here: (int)get_x(p), and by doing this we can display only integers on the fourth panel

richelbilderbeek:
Note: prefer using `static_cast<int>` over `(int)`, see [ES.49: If you must use a cast, use a named cast](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast) :-)

PR #512, opened 2022-03-31
Fix #455, #456 and #509

### LINKED ISSUE #455: Improve coordinates display
The coordinates displayed in the top-right corner are useful for development, but the last line (food item coordinates) looks a bit awkward: it's a long, single line, and half of it is out of the screen view! 

To make it clearer to read:
1. Slightly reduce the size of the font 
2. Split the text line in two, one for the x and one for the y

Hint: check out [SFML's tutorial](https://www.sfml-dev.org/tutorials/2.5/graphics-text.php) for examples of how text is drawn on screen, and modify something that looks like it in the code.

### LINKED ISSUE #509: Implement << operator for the player_state class
The class `player_state` does not have an `<<` operator yet.

Please add one, start by adding this test to the `player_state.cpp` file inside the `test_player_state()` function.

```
  // operator<<
  {
    std::stringstream s;
    const player_state ps = player_state::active;
    s << ps;
  }
  {
    const player_state ps = player_state::active;
    std::cout << ps;
  }
```

You can get inspired on how to implement the operator from class `sound_type`
`sound_type.h`
```
/// implement << operator
std::ostream& operator << (std::ostream& os, const sound_type& sound);
```
sound_type.cpp
```
std::ostream& operator << (std::ostream& os, const sound_type& sound)
{
    os << to_str(sound);
    return os;
}
```

EvoLandEco:
Closes #509

### DIFF
--- a/game_view.cpp
+++ b/game_view.cpp
@@ -285,13 +285,13 @@ void game_view::draw_player_coords() noexcept
     std::string str_player_coords;
     for(int i = 0; i != static_cast<int>(v_player.size()); i++) {
         player p = v_player[static_cast<unsigned int>(i)];
-        str_player_coords += "Player " + p.get_ID() + " x = " + std::to_string((int)get_x(p));
-        str_player_coords += "\nPlayer " + p.get_ID() + " y = " + std::to_string((int)get_y(p));
+        str_player_coords += "Player " + p.get_ID() + " x = " + std::to_string(static_cast<int>(get_x(p)));
+        str_player_coords += "\nPlayer " + p.get_ID() + " y = " + std::to_string(static_cast<int>(get_y(p)));
         str_player_coords += "\n\n";
     }
     food f = m_game.get_food()[0];
-    str_player_coords += "Food x = " + std::to_string((int)get_x(f));
-    str_player_coords += "\n         y = " + std::to_string((int)get_y(f));
+    str_player_coords += "Food x = " + std::to_string(static_cast<int>(get_x(f)));
+    str_player_coords += "\n         y = " + std::to_string(static_cast<int>(get_y(f)));
     str_player_coords += "\n\n";
 
     text.setString(str_player_coords);
--- a/player_state.cpp
+++ b/player_state.cpp
@@ -44,16 +44,12 @@ void test_player_state()
 
   #define FIX_ISSUE_509
   #ifdef FIX_ISSUE_509
-    // operator<<
-    {
-      std::stringstream s;
-      const player_state ps = player_state::active;
-      s << ps;
-    }
-    {
-      const player_state ps = player_state::active;
-      std::cout << ps;
-    }
+  // operator<<
+  {
+    std::stringstream s;
+    const player_state ps = player_state::active;
+    s << ps;
+  }
   #endif
   #endif
 }


game_view.cpp
```diff
@@ -278,19 +278,20 @@ void game_view::draw_player_coords() noexcept
 {
     sf::Text text;
     text.setFont(m_game_resources.get_font());
+    text.setCharacterSize(24);
```
richelbilderbeek, line 281: Hi @EvoLandEco, I appreciate this minor improvement and I will accept it. However, keep in mind that commits should ideally do exactly one thing. There are exceptions (it is called 'The Boy Scout Rule', where you leave the code a little cleaner than you found it, which usually applies to indentation). Here, however, you change the **behavior** of the program, and this may result in merge conflict with people that work on that text. 

But again, good stuff!

game_view.cpp
```diff
@@ -278,19 +278,20 @@ void game_view::draw_player_coords() noexcept
...
 
     // Concatenate player coordinates string
     std::vector<player> v_player = m_game.get_v_player();
     std::string str_player_coords;
     for(int i = 0; i != static_cast<int>(v_player.size()); i++) {
         player p = v_player[static_cast<unsigned int>(i)];
-        str_player_coords += "Player " + p.get_ID() + " x = " + std::to_string(trunc(get_x(p)));
-        str_player_coords += "\nPlayer " + p.get_ID() + " y = " + std::to_string(trunc(get_y(p)));
+        str_player_coords += "Player " + p.get_ID() + " x = " + std::to_string((int)get_x(p));
```
richelbilderbeek, line 288: I understand the thought process here and I could agree with a cast to int over a `trunc`.

However, the C++ Coreguideline [ES.49: If you must use a cast, use a named cast](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast) disagrees with you here. Due to this, I will ask you to improve this, as this helps you learn those C++ details better :-)

The reason why `(int)x` is a worse idea than `static_cast<int>(x)` is exactly because the latter is longer and ugly: usually casts are to be avoided.

Change this either back to `trunct` or use `static_cast<int>` and I will happily accept :+1:

player_state.cpp
```diff
@@ -17,6 +17,11 @@ std::string to_str(player_state this_player_state)
   }
 }
 
+std::ostream& operator << (std::ostream& os, const player_state& player)
```
richelbilderbeek, line 20: Brilliant! I could not have done this better :-)

player_state.cpp
```diff
@@ -39,5 +41,20 @@ void test_player_state()
     assert(to_str(player_state::stunned) == "stunned");
   }
   #endif // FIX_ISSUE_276
+
+  #define FIX_ISSUE_509
```
richelbilderbeek, line 45: After the test passes, you can remove the `#define`, `#ifndef` and `#ifdef` lines. It is optional as I use that for newbie Issues :-)

player_state.h
```diff
@@ -1,6 +1,8 @@
 #ifndef PLAYER_STATE_H
 #define PLAYER_STATE_H
 
+#include <iostream>
```
richelbilderbeek, line 4: I predict this include is the one that is the best of your knowledge. Pro-tip: use `#include <iosfwd>`. I will accept this regardless :-)

richelbilderbeek (changes requested):
Nice progress! I have some minor feedback on things I accept regardless and I am looking forward to see that back in future PRs :-) . There was 1 clear C++ Core Guideline violation, which I will ask you to fix: I link to the Core Guideline and give away the solution. I predict you enjoy that :-)

So, fix that and I'd be happy to accept the next PR! Yet again: good job!

game_view.cpp
EvoLandEco: I just got to know that my subsequent commits will be added to this pull request, so I decided to solve several issues in one PR, but of course I will try to avoid it next time

game_view.cpp
EvoLandEco: since trunc() does not work as intended, I would prefer to use static_cast<int> here, and thanks for the knowledge

richelbilderbeek: That is indeed the wiser choice :+1:

player_state.h
EvoLandEco: if include `iosfwd` instead of `iostream`, then `cout` will not be in `std::`, so that `std::cout << ps;` will fail. How to solve this issue?

EvoLandEco: I removed std::cout << ps, I assume this test is not important and for saving compiling time it would be better to stick to `iosfwd`

richelbilderbeek: @EvoLandEco my bad: in header files, use `iosfwd`, in implementation (`.cpp`) files, include `iostream`. This has to do with forward declarations (whatever those are :-) ).

The test you changed **is** important; it proes that a `player_state` can be streamed. No worries, I/you will create an Issue to put it back in :-)

EvoLandEco: I have put it back, with suggested `iosfwd` and `iostream` usage

richelbilderbeek: @EvoLandEco awesome :+1:

richelbilderbeek (commented):
Well done!

Issue #517, opened 2022-04-01
Implement fire rate for stun rocket


EvoLandEco:
There should also be a cool down time for the stun rocket

EvoLandEco:
Stun rocket belongs to the projectile class

EvoLandEco:
Shooting cooldown does not work on stun rocket at the moment.

Similar to #513 #518, we may add tests like this:
// Player enters cool down status after shooting stun rocket
  {
    game g;
    g.do_action(0, action_type::shoot_stun_rocket);
    g.tick();
    assert(g.get_player(0).is_cooling_down());
    g.do_action(0, action_type::shoot_stun_rocket);
    assert(!g.get_player(0).is_shooting_stun_rocket());
  }

  // Player can shoot stun rocket again after the cool down interval, player cannot shoot stun rocket during the interval
  {
    game g;
    g.do_action(0, action_type::shoot_stun_rocket);
    for (auto i = 0; i < projectile::m_fire_rate - 1; ++i)
      {
        g.tick();
        assert(g.get_player(0).is_cooling_down());
        g.do_action(0, action_type::shoot_stun_rocket);
        assert(!g.get_player(0).is_shooting_stun_rocket());
      }
    g.tick();
    assert(!g.get_player(0).is_cooling_down());
    g.do_action(0, action_type::shoot_stun_rocket);
    assert(g.get_player(0).is_shooting_stun_rocket());
  }

  // Player 0 enters cool down status will not affect other players
  {
    game g;
    g.do_action(0, action_type::shoot_stun_rocket);
    g.tick();
    assert(g.get_player(0).is_cooling_down());
    assert(!g.get_player(1).is_cooling_down());
    assert(!g.get_player(2).is_cooling_down());
  }
  #endif

Issue #516, opened 2022-04-01
Let each projectile type has its own fire rate


EvoLandEco:
IMO after the implementation of fire rate, each type of projectile should have its own fire rate

TheoPannetier:
I agree! But I also wonder about what the rest of the group think of this design decision.
@richelbilderbeek and @sebmader, this makes me think that we don't have a formal framework to discuss design decisions? We have the [Game Design](https://github.com/tresinformal/game/projects/6) kanban but don't really make use of it so far. Maybe there could be a way to exploit it to collectively discuss/validate/reject design decisions?

I'm thinking something like this:
1. Member has an idea for adding/changing something about game design
2. Member writes an issue about it, briefly exposes the proposition (like this issue) and possibly their motivations for it. Member adds label "Idea" to the issue, then assigns it to column "Ideas" in project "Game design"
3. Member signals the issue on Discord, possibly explains it for 5 min during a session, then the idea is submitted to vote. Issue is moved to "In-discussion".
4. When majority (or consensus?) is reached, the issue moves to "Agreed upon" and someone can then write a test for it and implement the proposition, as usual.

richelbilderbeek:
Nice discussion here! The informal flow is to let newbies implement their things, as those new things are usually good ideas. Sure, there is a risk that it gets removed in the future by voting (which has never happened until now). 

Also, when there is need for voting, I can always let that happen!

richelbilderbeek:
An easy solution: use a parameter for both the firing speeds. In that way, they can easily be set to equal values if desired. `game_options` is the best candidate at the moment, although that class should be split up as well (e.g. into `player_options` that contains all constants for a `player`)

sebmader:
Sounds like a fun idea!
To be honest, at first glance I thought that this is not the time to think about such a detail.
But I agree with the loose flow, so, let's take up ideas when they come to mind :)
If possible, the more senior people can take over the documenting of ideas on the respective board.
Not necessarily, so we vote about it before the idea is further developed, but to kind of keep track of our design decisions.

richelbilderbeek:
What about we vote next Wednesday on this? 

For example, at 18:15, each idea gets 3 mins max, after there will be vote. Team members can vote multiple times. Idea with most votes wins. In case of a draw, we vote with the top ideas only, with 1 vote per person. 

[...]

Issue #513, opened 2022-03-31
Add fire rate limit to projectile


EvoLandEco:
Press and hold the fire button will create a bunch of rockets which seems not to be the intended behavior. Fire rate limit is therefore important and it can be extended later as a part of the weapon upgrade system.

TheoPannetier:
Indeed, this is not intended behaviour, and as far as I remember no one has attempted to address it yet! Good luck ☘️

PR #518, opened 2022-04-01
Implement fire rate for projectile

EvoLandEco:
- [x] https://github.com/tresinformal/game/issues/513
- [ ] https://github.com/tresinformal/game/issues/516
- [ ] https://github.com/tresinformal/game/issues/517
- [ ] ...

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -116,10 +116,10 @@ void game::do_action(player& player, action_type action)
       }
       case action_type::shoot:
       {
-        if (player.is_calm_down() == false)
+        if (player.is_cool_down() == false)
         {
           player.shoot();
-          player.shoot_calm_down();
+          player.shoot_cool_down();
         }
         break;
       }
@@ -269,9 +269,9 @@ void game::tick()
   // Regenerate food items
   regenerate_food_items();
 
-  increment_shoot_calm_down_timers();
+  increment_shoot_cool_down_timers();
 
-  reset_shoot_calm_down_status();
+  reset_shoot_cool_down_status();
 
   // players that shoot must generate projectiles
   for (player &p : m_player)
@@ -361,23 +361,23 @@ void game::eat_food(food& f)
   f.reset_timer();
 }
 
-void game::increment_shoot_calm_down_timers()
+void game::increment_shoot_cool_down_timers()
 {
   for (player &p : m_player)
     {
-      p.increment_shoot_calm_down_timer();
+      p.increment_shoot_cool_down_timer();
     }
 }
 
-void game::reset_shoot_calm_down_status()
+void game::reset_shoot_cool_down_status()
 {
   for (player &p : m_player)
     {
       float player_shoot_fire_rate = projectile::m_fire_rate / p.get_shoot_fire_rate_multiplier();
-      if (p.is_calm_down() && p.get_shoot_calm_down_timer() >= player_shoot_fire_rate)
+      if (p.is_cool_down() && p.get_shoot_cool_down_timer() >= player_shoot_fire_rate)
         {
-          p.reset_shoot_calm_down_timer();
-          p.stop_shoot_calm_down();
+          p.reset_shoot_cool_down_timer();
+          p.stop_shoot_cool_down();
         }
     }
 }
--- a/game.cpp
+++ b/game.cpp
@@ -116,7 +116,11 @@ void game::do_action(player& player, action_type action)
       }
       case action_type::shoot:
       {
-        player.shoot();
+        if (player.is_calm_down() == false)
+        {
+          player.shoot();
+          player.shoot_calm_down();
+        }
         break;
       }
       case action_type::shoot_stun_rocket:
@@ -246,15 +250,16 @@ void game::tick()
   // For now only applies inertia
   apply_inertia();
 
-  //Move shelters
+  // Move shelters
   move_shelter();
 
-  //Actions issued by the players are executed
+  // Actions issued by the players are executed
   do_actions();
 
-  //Check and resolve wall collisions
+  // Check and resolve wall collisions
   resolve_wall_collisions();
 
+  // Increment timers of shoot calm down
   // Increment timers of all food elements
   increment_food_timers();
 
@@ -263,6 +268,10 @@ void game::tick()
 
   // Regenerate food items
   regenerate_food_items();
+
+  increment_shoot_calm_down_timers();
+
+  reset_shoot_calm_down_status();
 
   // players that shoot must generate projectiles
   for (player &p : m_player)
@@ -350,6 +359,27 @@ void game::eat_food(food& f)
     }
   f.set_food_state(food_state::eaten);
   f.reset_timer();
+}
+
+void game::increment_shoot_calm_down_timers()
+{
+  for (player &p : m_player)
+    {
+      p.increment_shoot_calm_down_timer();
+    }
+}
+
+void game::reset_shoot_calm_down_status()
+{
+  for (player &p : m_player)
+    {
+      float player_shoot_fire_rate = projectile::m_fire_rate / p.get_shoot_fire_rate_multiplier();
+      if (p.is_calm_down() && p.get_shoot_calm_down_timer() >= player_shoot_fire_rate)
+        {
+          p.reset_shoot_calm_down_timer();
+          p.stop_shoot_calm_down();
+        }
+    }
 }
 
 void game::increment_food_timers()
--- a/player.h
+++ b/player.h
@@ -85,7 +85,7 @@ public:
 
     bool is_shooting_stun_rocket() const noexcept { return m_is_shooting_stun_rocket; }
 
-    bool is_calm_down() const noexcept {return m_is_shoot_calm_down; }
+    bool is_cool_down() const noexcept {return m_is_shoot_cool_down; }
 
     ///Places a player to a given x,y poisition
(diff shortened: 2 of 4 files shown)


EvoLandEco:
> Hi @EvoLandEco, great to see you pick up this Issue! The first thing I looked for are the tests. There appear to be none. In that way, I will never be convinced this new code actually works. Also, it is likely the software architecture will be less good (and I already see an example of that). So, please add tests to convince me and the team this actually works 👍
> 
> Due to this, I have not looked at the rest of the code, as I feel it too likely to change a bit, due to the addition of one or more tests.
> 
> If you need help writing tests, I (and @TheoPannetier is also getting good at this) would be hapy to help!

Thanks @richelbilderbeek @TheoPannetier for explaining the test-driven development to me, my workflow has to change because I always write post-hoc tests. I will try to add some tests now, hope it won't be too late 🤣

EvoLandEco:
I hope the tests can cover most of the problem. I would like to hear your opinion😀

game.cpp
```diff
@@ -352,6 +361,30 @@ void game::eat_food(food& f)
...
+        p.increment_shoot_cool_down_timer();
+      }
+    }
+}
+
+void game::reset_shoot_cool_down_status()
+{
+  for (player &p : m_player)
+    {
+      float player_shoot_fire_rate = projectile::m_fire_rate / p.get_shoot_fire_rate_multiplier();
```
TheoPannetier, line 379: I don't understand why the fire rate should be player-specific here... My guess is that you have future plans for mechanisms that will allow players to shoot more or less fast after eating some food, going in a specific part of the terrain or something else like this? If so, this is great and I would encourage you to do it, but for now this adds complexity that is unnecessary to solve the issue (#513).
For now, I would advise to keep this idea aside, and if you have a greater plan for changing the fire rate based on a player's condition, write spearate issues and tests for it first 😉

game.cpp
```diff
@@ -717,6 +750,39 @@ void test_game() //!OCLINT tests may be many
...
+    assert(g.get_player(0).is_shoot_cool_down());
+    g.do_action(0, action_type::shoot);
+    assert(!g.get_player(0).is_shooting());
+  }
+
+  // Player can shoot again after the interval
+  {
+    game g;
+    g.do_action(0, action_type::shoot);
+    for (auto i = 0; i <= projectile::m_fire_rate / g.get_player(0).get_shoot_fire_rate_multiplier(); ++i)
```
TheoPannetier, line 767:
```suggestion
    for (auto i = 0; i <= projectile::m_fire_rate; ++i)
```

game.cpp
```diff
@@ -717,6 +750,39 @@ void test_game() //!OCLINT tests may be many
       }
   }
 
+  // Player enters cool down status after one shooting action
```
TheoPannetier, line 753: Excellent, excellent work on these three tests!

game.cpp
```diff
@@ -717,6 +750,39 @@ void test_game() //!OCLINT tests may be many
...
+    assert(!g.get_player(0).is_shooting());
+  }
+
+  // Player can shoot again after the interval
+  {
+    game g;
+    g.do_action(0, action_type::shoot);
+    for (auto i = 0; i <= projectile::m_fire_rate / g.get_player(0).get_shoot_fire_rate_multiplier(); ++i)
+      {
+        g.tick();
```
TheoPannetier, line 769: I feel a welcome (but optional) bonus would be to check that the player still cannot shoot during any of these ticks (except the last one I think?)

game.cpp
```diff
@@ -116,7 +116,11 @@ void game::do_action(player& player, action_type action)
       }
       case action_type::shoot:
       {
-        player.shoot();
+        if (player.is_shoot_cool_down() == false)
```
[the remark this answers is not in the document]
TheoPannetier: If you do decide to go with my suggestion and change the name, don't forget to also do the same for all corresponding variables and functions 😉

player.h
```diff
@@ -109,6 +111,27 @@ class player
...
+
+    /// Get the current shoot calm down timer value
+    int get_shoot_cool_down_timer() {return m_shoot_cool_down_timer; }
+
+    /// Get the player's shoot fire rate multiplier
+    float get_shoot_fire_rate_multiplier() {return m_shoot_fire_rate_multiplier; }
+
+    /// Change the player's shoot fire rate multiplier
+    void change_shoot_fire_rate_multiplier(int m) noexcept {m_shoot_fire_rate_multiplier = m; }
+
```
TheoPannetier, line 134: Good job at choosing evocative action verbs for the function names!

game.cpp
```diff
@@ -264,6 +269,10 @@ void game::tick()
   // Regenerate food items
   regenerate_food_items();
 
+  increment_shoot_cool_down_timers();
+
+  reset_shoot_cool_down_status();
+
```
TheoPannetier, line 275: Examplary work on following [F.1](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-package)! Good job!

game.cpp
TheoPannetier, line 753:
```suggestion
  // (513) Player enters cool down status after one shooting action
```

We try to refer to issues names in the tests :)
This makes it easier to cross-reference issues and test if someone doesn't understand why it's there later on

TheoPannetier (changes requested):
Excellent work here, almost impeccable! I'm impressed with how quickly you picked up the expected structure of the game logic, the content of tests and what makes good variable/function names.
My only mandatory request is that you remove player-specific fire rates (for now), as this does not help solving the issue at hand. 
I have made other suggestions to improve the PR, but feel free to reject any of them.

Also, since this appears to be only about #513, could you edit the PR's description to remove mentions to other issues? Just for archiving purposes :)

game.cpp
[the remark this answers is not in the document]
EvoLandEco: Seems like I misunderstood how the projectiles are defined.  I thought that stun rocket was defined differently from the normal projectile, that is why I used shoot_cool_down but not cooling_down. I will refactor all of them

game.cpp
```diff
@@ -116,7 +116,11 @@ void game::do_action(player& player, action_type action)
       }
       case action_type::shoot:
       {
-        player.shoot();
+        if (player.is_shoot_cool_down() == false)
+        {
+          player.shoot();
+          player.shoot_cool_down();
```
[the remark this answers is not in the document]
EvoLandEco: That's true, player.trigger_cool_down() is a good choice, thanks for the suggestion

game.cpp
```diff
@@ -246,15 +250,16 @@ void game::tick()
...
 
-  //Actions issued by the players are executed
+  // Actions issued by the players are executed
   do_actions();
 
-  //Check and resolve wall collisions
+  // Check and resolve wall collisions
   resolve_wall_collisions();
 
+  // Increment timers of shoot calm down
```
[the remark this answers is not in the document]
EvoLandEco: Yes, the comments were misplaced. Will fix them

game.cpp
EvoLandEco: I do plan to implement a weapon upgrade system in the near future, but I will surely break down the whole thing into several issues and solve them one at a time.

game.cpp
EvoLandEco: I will follow the convention of the code here

game.cpp
EvoLandEco: And one thing that makes the code less robust here is that I divided the fire rate by the player-specific multiplier, this will probably generate a float, which may introduce issues in later development since ticks should be integer. If I implement player-specific fire rate again, I may need to round the value, or use addition instead of multiplication or division? I think this would be interesting if we have time to discuss the game design in a later meeting.

game.cpp
EvoLandEco: I was thinking about it, I did not implement this because I believe the case was already covered in the first test. But of course, adding this will further strengthen the safe net, I will add it to the test

game.cpp
```diff
@@ -717,6 +750,39 @@ void test_game() //!OCLINT tests may be many
...
+
+  // Player 0 enters cool down status will not affect other players
+  {
+    game g;
+    g.do_action(0, action_type::shoot);
+    g.tick();
+    assert(g.get_player(0).is_shoot_cool_down());
+    assert(!g.get_player(1).is_shoot_cool_down());
+    assert(!g.get_player(2).is_shoot_cool_down());
+  }
```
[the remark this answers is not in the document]
EvoLandEco: The mechanism is implemented within the player class, so it is not likely that one player object's behavior will affect others. But I still prefer to keep this test just in case

game.cpp
richelbilderbeek: @EvoLandEco in the future, there will be a `player_options` class, containing all constants a player has, such as, for example, the firing rate. I feel these can easily differ per player :+1:

game.cpp
richelbilderbeek: @EvoLandEco the rule is that a test should break the code. If it did not, the test can be deleted. And yes, there will be tests that overlap in testing the same thing. You may delete some obsolete tests, just make sure it is reviewed :-)

richelbilderbeek (approved):
Excellent work, also thanks to @TheoPannetier's comments. Sorry for the dull code review :-)

Let's get this to `develop` :+1:

Issue #522, opened 2022-04-06
Save and load a key_action_map


richelbilderbeek:
A `key_action_map` is part of the options of a game. It needs to be saved, so that a user can store his/her key setting. Allow a `key_action_map` to be saved and loaded.

PR #523, opened 2022-04-06
Progress in saving a `key_action_map`, #522

richelbilderbeek:
Here is some stub code to implement #522 as criticized during the mob programming by me at 2022-04-06. It is not complete, but the tests and function declarations are there.

Also the excess output (#526) is removed :-)

### DIFF
--- a/key_action_map.cpp
+++ b/key_action_map.cpp
@@ -186,17 +186,33 @@ sf::Keyboard::Key to_sfml_key(const std::string& s)
 {
   if (s == "A") return sf::Keyboard::Key::A;
 
+  if (s != " ") std::clog << "UNKNOWN KEY: " << s << '\n';
 
   assert(s == " "); // Or the key has not been encoded yet
   return sf::Keyboard::Key::Space;
 
 }
 
+std::string to_str(const sf::Keyboard::Key key)
+{
+  switch (key)
+  {
+    case sf::Keyboard::Key::A: return "A";
+    case sf::Keyboard::Key::D: return "D";
+    default: break;
+  }
+
+  // Or maybe the key is not present yet in the
+  // switch statement above :-)
+  assert(key == sf::Keyboard::Key::Space);
+  return " ";
+}
+
 std::string to_str(const key_action_map& kam) noexcept
 {
   std::stringstream s;
   for (const auto& p: kam.get_raw_map()) {
-    s << p.first << " " << p.second << " ";
+    s << to_str(p.first) << " " << p.second << " ";
   };
   std::string t = s.str();
   // Remove the space at the end
@@ -206,19 +222,34 @@ std::string to_str(const key_action_map& kam) noexcept
 
 std::istream& operator>>(std::istream& is, key_action_map& kam)
 {
+  std::string action = "dummy"; // In istream for readability
+
+  // Go through each key action pair, check if the order is correct
   std::string key_to_go_left;
+  is >> key_to_go_left >> action;
+  assert(to_str(action_type::turn_left) == action);
+
   std::string key_to_go_right;
-  std::string key_to_accelerate_forward;
-  std::string key_to_accelerate_backward;
-  std::string key_to_shoot;
+  is >> key_to_go_right >> action;
+  assert(to_str(action_type::turn_right) == action);
+
   std::string key_to_stun;
-  is >> key_to_go_left
-    >> key_to_go_right
-    >> key_to_accelerate_forward
-    >> key_to_accelerate_backward
-    >> key_to_shoot
-    >> key_to_stun
-  ;
+  is >> key_to_stun >> action;
+  assert(to_str(action_type::shoot_stun_rocket) == action);
+
+  std::string key_to_shoot;
+  is >> key_to_shoot >> action;
+  assert(to_str(action_type::shoot) == action);
+
+  std::string key_to_accelerate_backward;
+  is >> key_to_accelerate_backward >> action;
+  assert(to_str(action_type::accelerate_backward) == action);
+
+  std::string key_to_accelerate_forward;
+  is >> key_to_accelerate_forward >> action;
+  assert(to_str(action_type::accelerate_forward) == action);
+
+
   const key_action_map m(
     to_sfml_key(key_to_go_left),
     to_sfml_key(key_to_go_right),
--- a/key_action_map.cpp
+++ b/key_action_map.cpp
@@ -168,11 +168,6 @@ sf::Keyboard::Key get_stun_key(const key_action_map& m)
   return m.to_key(action_type::shoot_stun_rocket);
 }
 
-bool exists_file (const std::string& name) {
-    std::ifstream f(name.c_str());
-    return f.good();
-}
-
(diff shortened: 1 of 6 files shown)


key_action_map.cpp
```diff
@@ -163,6 +168,77 @@ sf::Keyboard::Key get_stun_key(const key_action_map& m)
...
+
+void save_to_file(const key_action_map& kam, const std::string& filename)
+{
+  std::ofstream file(filename);
+  file << kam;
+}
+
+sf::Keyboard::Key to_sfml_key(const std::string& s)
+{
+  if (s == "A") return sf::Keyboard::Key::A;
```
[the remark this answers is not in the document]
richelbilderbeek: Hi @EvoLandEco, indeed, this is a clumsy incomplete function. The C++20/C++24 equivalent of the R `eval` is called 'reflection'. Because we do not use C++20/C++24 (yet), this clumsy way is the only work forward.

There will be smarter ways (using a look-up `std::map`), which will be a future junior Issue :-)

key_action_map.cpp
```diff
@@ -163,6 +168,77 @@ sf::Keyboard::Key get_stun_key(const key_action_map& m)
   return m.to_key(action_type::shoot_stun_rocket);
 }
 
+key_action_map load_kam(const std::string& filename)
```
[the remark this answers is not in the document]
richelbilderbeek: I agree that this can be a problem. What about we create an Issue for that?

key_action_map.cpp
```diff
@@ -163,6 +168,77 @@ sf::Keyboard::Key get_stun_key(const key_action_map& m)
...
+{
+  std::ifstream f(filename);
+  key_action_map m;
+  f >> m;
+  return m;
+}
+
+void save_to_file(const key_action_map& kam, const std::string& filename)
```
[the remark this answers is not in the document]
EvoLandEco: Ah I see it here, and I think that's the same thing as `to_str()`, which has been overloaded several times. I now understand the reasoning and I agree with you : ）

key_action_map.cpp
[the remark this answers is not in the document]
EvoLandEco: Good to know that C++ finally has this quality of life feature, though I see many C++ programmers questioning the efficiency and safety issues of doing this

key_action_map.cpp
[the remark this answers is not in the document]
EvoLandEco: Created #529

key_action_map.cpp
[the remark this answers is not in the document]
EvoLandEco: Created #530

key_action_map.cpp
```diff
@@ -310,4 +386,22 @@ void test_key_action_map()//!OCLINT tests can be many
...
+    std::stringstream s;
+    s << kam;
+  }
+  {
+    const key_action_map kam = get_player_1_kam();
+    const std::string filename = "test.txt";
+    save_to_file(kam, filename);
+    const key_action_map map_again = load_kam(filename);
+    assert(kam == map_again);
+    // TODO: delete temporary file
```
EvoLandEco, line 403: Here we need another issue to remind us to delete this temporary file

key_action_map.cpp
```diff
@@ -310,4 +386,22 @@ void test_key_action_map()//!OCLINT tests can be many
...
+    const key_action_map kam = get_player_1_kam();
+    std::stringstream s;
+    s << kam;
+  }
+  {
+    const key_action_map kam = get_player_1_kam();
+    const std::string filename = "test.txt";
+    save_to_file(kam, filename);
+    const key_action_map map_again = load_kam(filename);
+    assert(kam == map_again);
```
EvoLandEco, line 408: This test is failing, need to figure out what is wrong here

key_action_map.cpp
```diff
@@ -163,6 +168,82 @@ sf::Keyboard::Key get_stun_key(const key_action_map& m)
...
+  assert(s == " "); // Or the key has not been encoded yet
+  return sf::Keyboard::Key::Space;
+
+}
+
+std::string to_str(const key_action_map& kam) noexcept
+{
+  std::stringstream s;
+  for (const auto& p: kam.get_raw_map()) {
+    s << p.first << " " << p.second << " ";
```
EvoLandEco, line 204: Seems like `std::maps` sorts its elements in lexicographic order (alphabetic order) according to the first key. This loop will not generate a string based on the insertion order.

EvoLandEco: `0 turn_left 3 turn_right 4 shoot_stun_rocket 16 shoot 18 accelerate_backward 22 accelerate_forward`
This is the actual order in a saved file

key_action_map.cpp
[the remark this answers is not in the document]
EvoLandEco: And also: shouldn't it be `if (s == "3") return sf::Keyboard::Key::A;`?
In test.txt I see `3 turn_left`, but not `A turn_left`. I guess the `<<` operator overloaded for sfml::Keyboard will convert it to a digit which is the index of the key?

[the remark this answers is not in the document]
richelbilderbeek: > I guess the << operator overloaded for sfml::Keyboard will convert it to a digit which is the index of the key?

I guess it does. For now. Some future junior Issues (where one of write the tests) should fix that :-)

key_action_map.cpp
richelbilderbeek: Ah, my bad, thanks for rejecting this PR: I would have messed up develop :-)

key_action_map.cpp
richelbilderbeek: I'll take a look :-)

richelbilderbeek:
Hi @EvoLandEco, thanks for the first review round. I'll make sure `develop` does not break, as well as taking a closer look at a comment :-)

key_action_map.cpp
EvoLandEco: You disabled the tests, so develop won't break if we merge this PR. I was playing a bit with the tests to see whether your code works, if we are going to leave the issues to the new devs, then I will disable them again

richelbilderbeek: Ah, I guess you found out why I disabled that test :-)

key_action_map.cpp
EvoLandEco: @richelbilderbeek I've reverted my last commit on your branch, I'll switch to my own branch, let's open an issue for this. But can we approve this PR without functioning tests?

EvoLandEco:
Would it be a good idea if I introduce an ini parser/generator? I found a interesting one here https://github.com/mcmtroffaes/inipp/

For the purpose of writing/reading the key action maps and maybe later the player settings?

key_action_map.cpp
richelbilderbeek: @EvoLandEco yes, one can approve of the tests :-) . Good tests are hard to write :+1:

richelbilderbeek:
This PR is sitting idle for too lomg now. I feel it has been reviewed enough, I will merge it and even take responsibility for the merge :-)

PR #525, opened 2022-04-11
Rework the movement system

### LINKED ISSUE #524: [Game Play] Rework the movement system
I feel like the current movement control is not smooth. There is a test of whether inertia slows down players, though not robust. IMO the movement system can to be optimized, and more robust tests should be added.

Despite of fixing the bugs, I would also like to make the movement system more flexible and player-friendly, the new system would be able to provide a smoother control of the protagonists, and to allow for further in-depth optimization.

I propose to give the `player` class the following parameters:

- `player_max_speed_forward` the maximum speed when moving forward
- `player_max_speed_backward` the maximum speed when moving backward
- `player_acceleration_forward` the acceleration forward when we press and hold the move forward key
- `player_acceleration_backward` the acceleration backward when we press and hold the move backward key
- `player_deceleration_forward` the deceleration due to inertia when moving forward, if we do not press the keys to move forward or backward
- `player_deceleration_backward` the deceleration due to inertia when moving backward, if we do not press the keys to move forward or backward

By doing this the players would be able accelerate and decelerate naturally and differently between forward and backward directions.

`action_type` class will be modified accordingly:

`action_type::brake` will be removed
`action_type::accelerate` will be changed to `action_type::accelerate_forward`
`action_type::acc_backward` will be changed to `action_type::accelerate_backward`

I also propose to give the `player` class a member variable of action flag, which indicates what the player is doing during the current tick, this enables us to do many things in the future, and at the moment, the ability to allow `game::apply_inertia()` to judge whether the player is pressing the keys to move forward or backward. If so, `game::apply_inertia()` should not work, and if not, the players should auto-decelerate due to the inertia.

I don't think `brake` is needed, firstly we already have inertia (with value large enough) which behaves as auto-brake, secondly this makes the control more difficult and less enjoyable, unless the effect of inertia is really small that we need to brake manually, and we want to make a relatively hard-core game.

Please advice me if there is a better way to implement this idea, and comment if you agree/disagree with me😀

EvoLandEco:
fixes #524 
To locate the tests please search FIX_ISSUE_524
I would like to recommend you to build and try the new movement to see whether it is a good idea


EvoLandEco:
To locate the tests please search `FIX_ISSUE_524`

game.cpp
```diff
@@ -92,26 +94,35 @@ void game::do_action(player& player, action_type action)
...
-        player.accelerate();
-        break;
-      }
-      case action_type::brake:
+      case action_type::accelerate_forward:
       {
-        player.brake();
+        if (player.get_speed() >= 0) {
+          player.accelerate_forward();
+          player.set_action(action_type::accelerate_forward);
```
richelbilderbeek, line 110: I am unsure if you need to do both modify the player, as well as setting the action (so that the player will do the action at the next tick). Maybe this is a duplicate of things happening that needs to be cleaned in the future, but for now, I'll accept as it passes all tests :-)

game.cpp
```diff
@@ -844,29 +890,6 @@ void test_game() //!OCLINT tests may be many
     g.tick();
     assert(g.get_n_ticks() - before == 1);
   }
-  // inertia  slows down players
```
richelbilderbeek, line 847: Inertia is a useful feature. Why is this removed? Or is it moved somewhere else?

game.cpp
```diff
@@ -1808,6 +1844,83 @@ void test_game() //!OCLINT tests may be many
...
+            assert(g.get_player(i).get_action() == action_type::accelerate_forward);
+        }
+        g.reset_player_action();
+        for (auto i = 0; i < static_cast<int>(g.get_v_player().size()); ++i)
+        {
+            assert(g.get_player(i).get_action() == action_type::none);
+        }
+    }
+
+    // apply_inertia() works properly
```
richelbilderbeek, line 1884: Ah, there is the inertia. This is a looooong test, it used to be shorter. I would enjoy it to be short again. If you test two different things, just use two different tests :-)

richelbilderbeek (changes requested):
This is a BIG Pull Request, but I guess that is what enthusiasm looks like :-)

I enjoy the ruthlessness in it (i.e. doing many things without discussing with the team) and I think that is preferable over always discussing every tiny detail before doing something.

All I need to hear: if the team misses the option of braking (i.e. the feature you removed :-) ), will you put it back in? If yes, I will happily accept!

richelbilderbeek:
BTW, I put some other comments at [https://github.com/tresinformal/game/issues/524#issuecomment-1095206050](https://github.com/tresinformal/game/issues/524#issuecomment-1095206050), but let's do those in a next PR :-)

TheoPannetier:
Hey, sorry but I won't find much time for reviewing PR's in the next three weeks, I'll have to un-assign myself from here 😬
May I suggest @sebmader as a reviewer? He's well acquainted with the movement system :)

action_type.cpp
```diff
@@ -9,17 +9,16 @@ void test_action_type()
   {
     assert(to_str(action_type::turn_left) == "turn_left");
     assert(to_str(action_type::turn_right) == "turn_right");
-    assert(to_str(action_type::accelerate) == "accelerate");
-    assert(to_str(action_type::brake) == "brake");
-    assert(to_str(action_type::acc_backward) == "acc_backward");
+    assert(to_str(action_type::accelerate_forward) == "accelerate_forward");
+    assert(to_str(action_type::accelerate_backward) == "accelerate_backward");
```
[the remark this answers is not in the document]
EvoLandEco: @richelbilderbeek given the current parameter set there is no need for `brake`. But when we use some slightly different parameter sets things could be much different. I won't disagree if the team's decision is to put brake back in, and I believe that the new movement system would not conflict with `brake`, just add two new parameters to decide how fast a player can brake in both directions. 

IMO the thing that really matters is how we set those parameters, and this may result in completely different tactile sensations of manipulating. But of course, this is still really far from being seriously considered.

In fact I see a draft pull request as a prototype that is still in progress and prone to change, so I am glad to hear the opinions from you and the entire team. I think that real code may inspire people more, that is why I opened this draft. I really like to make my idea come true ASAP, but I like more to make the implementation even better. So, just like before, please comment if you have any questions, and if there is a need to vote please do so😀

game.cpp
EvoLandEco: `set_action()` here will just set a player object's private member `m_action` to `action_type::accelerate_forward`, this will not let the player to do this action at the next tick. It will just give a player object a sign/flag. I do this because I want to detect what action the player is executing at the current tick, to be more specific, I want to detect what key is being pressed at the current tick.

Let's say if `W` is being pressed at the current tick, then the player 0 should accelerate, and `game::apply_inertia()` should be able to detect this fact and thus it won't make the player to decelerate due to the inertia which counteracts with the key `W` being pressed.

EvoLandEco:
> BTW, I put some other comments at [#524 (comment)](https://github.com/tresinformal/game/issues/524#issuecomment-1095206050), but let's do those in a next PR :-)

Yeah, finally, it is time to make `player_options` come true

EvoLandEco:
> Hey, sorry but I won't find much time for reviewing PR's in the next three weeks, I'll have to un-assign myself from here 😬 May I suggest @sebmader as a reviewer? He's well acquainted with the movement system :)

Thanks for your suggestion😀, good luck!

game.cpp
EvoLandEco: The test is long because I wish to test two things that one happens after another. 1) When there is key input of move forward or backward, `apply_inertia()` does nothing; 2) When there is no such input at the next tick, `apply_inertia()` decelerates players. I removed some safety nets (aka redundant tests) to make it shorter : )

game.cpp
EvoLandEco: Maybe we need to refactor this method to make it self-explained? Something like `set_action_flag()`

action_type.cpp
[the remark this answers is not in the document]
EvoLandEco: However if the new system is adopted, we need to find a new key and map it to `brake`, is it the preferable way?

game.cpp
richelbilderbeek: Excellent, I justed wanted to know if you gave it any thought or just randomly did stuff. You have had an idea what you were doing. That is enough for me! Refactoring the method to be more self-explanatory is great, especially after merging this PR to develop :innocent:

action_type.cpp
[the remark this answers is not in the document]
richelbilderbeek: Adding an action, or a key mapped to an action, back is trivial, just go ahead on with your works of wonder :+1:

EvoLandEco:
```
Run ./game_on_gha
game_on_gha: game.cpp:1919: void test_game(): Assertion `abs(g.get_player(0).get_max_speed_backward()) >= g.get_player(0).get_deceleration_backward()' failed.
/home/runner/work/_temp/a4325ad5-43ab-4c38-af96-bc81e1ea32eb.sh: line 1:  4547 Aborted                 (core dumped) ./game_on_gha
Error: Process completed with exit code [13](https://github.com/tresinformal/game/runs/5992780621?check_suite_focus=true#step:8:13)4.
```

@richelbilderbeek strange that assertion fails on GitHub Action but passes locally

richelbilderbeek:
@EvoLandEco,

> @richelbilderbeek strange that assertion fails on GitHub Action but passes locally

```
Run ./game_on_gha
game_on_gha: game.cpp:1919: void test_game(): Assertion `abs(g.get_player(0).get_max_speed_backward()) >= g.get_player(0).get_deceleration_backward()' failed.
/home/runner/work/_temp/a4325ad5-43ab-4c38-af96-bc81e1ea32eb.sh: line 1:  4547 Aborted                 (core dumped) ./game_on_gha
Error: Process completed with exit code [13](https://github.com/tresinformal/game/runs/5992780621?check_suite_focus=true#step:8:13)4.
```

Yup, too bad (yes, I really mean 'bad' here :-) ) your operating system does not detect this access violation. It is a property of access violations to be triggered irregularly. 

In our context, this is usually caused bu `v[x]`, where `v` is a `std::vector`, and `x` is an index that is not present in that vector. In the context of the error message, `g.get_player(0)` is probably the cause, where the player at index `0` is absent.

Tips: add `asserts` (e.g. `assert(index >= 0);` and `assert(index < static_cast<int>(v.size());` to document your assumptions (see [C++ Core Guideline I.6: Prefer Expects() for expressing preconditions](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-pre), read `assert` instead of `Expects` :-) )

EvoLandEco:
> @EvoLandEco,
> 
> > @richelbilderbeek strange that assertion fails on GitHub Action but passes locally
> 
> ```
> Run ./game_on_gha
> game_on_gha: game.cpp:1919: void test_game(): Assertion `abs(g.get_player(0).get_max_speed_backward()) >= g.get_player(0).get_deceleration_backward()' failed.
> /home/runner/work/_temp/a4325ad5-43ab-4c38-af96-bc81e1ea32eb.sh: line 1:  4547 Aborted                 (core dumped) ./game_on_gha
> Error: Process completed with exit code [13](https://github.com/tresinformal/game/runs/5992780621?check_suite_focus=true#step:8:13)4.
> ```
> 
> Yup, too bad (yes, I really mean 'bad' here :-) ) your operating system does not detect this access violation. It is a property of access violations to be triggered irregularly.
> 
> In our context, this is usually caused bu `v[x]`, where `v` is a `std::vector`, and `x` is an index that is not present in that vector. In the context of the error message, `g.get_player(0)` is probably the cause, where the player at index `0` is absent.
> 
> Tips: add `asserts` (e.g. `assert(index >= 0);` and `assert(index < static_cast<int>(v.size());` to document your assumptions (see [C++ Core Guideline I.6: Prefer Expects() for expressing preconditions](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-pre), read `assert` instead of `Expects` :-) )

I don't quite understand this violation. Doesn't vector index in C++ start from 0? According to my debug console, player 0 did exist and `Assertion abs(g.get_player(0).get_max_speed_backward()) >= g.get_player(0).get_deceleration_backward()` shouldn't fail because `abs(g.get_player(0).get_max_speed_backward())` returned 0.5 and `g.get_player(0).get_deceleration_backward()` returned 0.1. 

I cannot reproduce any error locally, I also tried to change `g.get_player(1)` to `g.get_player(0)`  in other tests, but they did not trigger the same error on GitHub Action.

richelbilderbeek:
Hi @EvoLandEco, the computer is not making things up. Indeed C++ vectors start at zero and if the things you claim are true, it must be something else. 

Congratulations with your first access violation, I predict we'll solve it tonight in the first half together. Would you mind if I and another team member hunt for the problem the first 30 mins?

EvoLandEco:
Hi Richel,

I have an R&O meeting which might end at 17:30, I may not be able to attend in time, please do so even if I am absent.

Tianjian
From mobile phone
________________________________

EvoLandEco:
@richelbilderbeek this bug may be due to an inconsistent behavior insides `game::tick()`. I set up a ubuntu subsystem on my laptop to be able to reproduce this error. 

The only difference according to two debug consoles on two platforms is the variable `after`, if you set a break point at line 1936 (please pull all the changes). On windows `after` is `-0.4`, this is the expected behavior and is what `apply_inertia()` should do. On Linux `after` is `0`, can we assume that `game::tick()` or something within it behaves differently across the platforms?

EvoLandEco:
@richelbilderbeek @ollyturner the problem is the same: on Linux we should use `std::abs()` instead of `abs()`, or a double will be coerced into an int. Now all tests are passing

player.cpp
```diff
@@ -82,7 +82,7 @@ void player::decelerate() noexcept
     }
     else if(m_player_speed < 0.0)
     {
-        if(abs(m_player_speed) >= m_player_deceleration_backward)
+        if(std::abs(m_player_speed) >= m_player_deceleration_backward)
```
EvoLandEco, line 85: abs() is designed for int while std::abs() can deal with float or double

EvoLandEco: https://stackoverflow.com/questions/21392627/abs-vs-stdabs-what-does-the-reference-say
link that might help

.gitignore
```diff
@@ -56,3 +56,4 @@ game_on_travis
 gprof.log
 game_on_gha
 .DS_Store
+vcpkg/
```
richelbilderbeek, line 59: Well done, thanks for keeping the repo clean :-)

PR #541, opened 2022-05-11
key-action-map of player 1 can be saved and loaded, progress #522

richelbilderbeek:
This is some progress on #522: now the key-action-map of player 1 can be saved and loaded. For other players, this still fails, due to, for examples, arrow keys. There are helpful `assert`s for that, so that anyone can continue easily.

### DIFF
--- a/key_action_map.cpp
+++ b/key_action_map.cpp
@@ -472,7 +472,7 @@ void test_key_action_map()//!OCLINT tests can be many
     std::stringstream s;
     s << kam;
   }
-  // 522. Can save and load a key_action_map
+  // 522. Can save and load a key_action_map for player 1
   {
     const key_action_map kam = get_player_1_kam();
     const std::string filename = "test.txt";


key_action_map.cpp
```diff
@@ -185,6 +185,31 @@ void save_to_file(const key_action_map& kam, const std::string& filename)
...
+  if (s == "Q") return sf::Keyboard::Key::Q;
+  if (s == "R") return sf::Keyboard::Key::R;
+  if (s == "S") return sf::Keyboard::Key::S;
+  if (s == "T") return sf::Keyboard::Key::T;
+  if (s == "U") return sf::Keyboard::Key::U;
+  if (s == "V") return sf::Keyboard::Key::V;
+  if (s == "W") return sf::Keyboard::Key::W;
+  if (s == "X") return sf::Keyboard::Key::X;
+  if (s == "Y") return sf::Keyboard::Key::Y;
+  if (s == "Z") return sf::Keyboard::Key::Z;
```
TheoPannetier, line 212: There is probably a very smart alternative to these 26 lines of code, but I enjoy the safety and readability of this option 👍

key_action_map.cpp
```diff
@@ -417,22 +466,40 @@ void test_key_action_map()//!OCLINT tests can be many
...
+  // 522. Can save and load a key_action_map
+  {
+    const key_action_map kam = get_player_2_kam();
+    const std::string filename = "test.txt";
+    save_to_file(kam, filename);
+    const key_action_map map_again = load_kam(filename);
+    assert(kam == map_again);
+    std::remove(filename.c_str());
+  }
+  // 522. Can save and load a key_action_map
```
TheoPannetier, line 494: Does this title need to be repeated for every player?
```suggestion

```

key_action_map.cpp
```diff
@@ -417,22 +466,40 @@ void test_key_action_map()//!OCLINT tests can be many
...
     const key_action_map kam = get_player_1_kam();
     const std::string filename = "test.txt";
     save_to_file(kam, filename);
     const key_action_map map_again = load_kam(filename);
     assert(kam == map_again);
-    // TODO: delete temporary file
+    std::remove(filename.c_str());
+  }
+  #ifdef FIX_ISSUE_522
+  // 522. Can save and load a key_action_map
```
TheoPannetier, line 485:
```suggestion

```

TheoPannetier (approved):
Nice progress! I would enjoy if you could report what's left to do for this issue in the description of #522 or its thread and why it doesn't work yet; but I'll accept this PR regardless.

key_action_map.cpp
richelbilderbeek: Unsure if it should. I enjoy every section of code to have its own heading and if two pieces of code do the same, the heading will be the same. But as it is more of a personal preference, I'd be wiling to remove it or give it another text.

key_action_map.cpp
richelbilderbeek: It is clumsy indeed. If someone wants to practice making code smarter, I think this is a fun section to work on :-)

key_action_map.cpp
TheoPannetier:
```suggestion
  // 522. Can save and load a key_action_map for player 3
```

key_action_map.cpp
TheoPannetier:
```suggestion
  // 522. Can save and load a key_action_map for player 2
```

key_action_map.cpp
```diff
@@ -417,22 +466,40 @@ void test_key_action_map()//!OCLINT tests can be many
...
-
-//#define ISSUE_522
-#ifdef ISSUE_522
+  // 522. operator<< for key_action_map
   {
     const key_action_map kam = get_player_1_kam();
     std::stringstream s;
     s << kam;
   }
+  // 522. Can save and load a key_action_map
```
TheoPannetier, line 475:
```suggestion
  // 522. Can save and load a key_action_map for player 1
```

key_action_map.cpp
TheoPannetier: If it's all the same, I added an extra bit to be more specific ;)

richelbilderbeek:
@TheoPannetier : done! Thanks for the help, we properly attributed the glory to you :-)

Issue #545, opened 2022-05-12
[Menu] Only open a window when the corresponding view is used


TheoPannetier:
Since #488, we can now switch from one view to the next, e.g. from game to menu.
However, under the current implementations of the `_view` classes, all windows for each view are drawn at the beginning of main, even if not used yet, which clutters the screen. For example, starting with the (default) game view, we get this:

![pic_544](https://user-images.githubusercontent.com/34482571/168157755-479646b7-2221-48c4-bf70-36d3180e8cb1.PNG)

Then if I switch to menu, the game window is closed and the menu view is drawn in the menu window. 
~~One could probably close and re-open windows whenever the corresponding view is called, but it would be more elegant to have a single window for the game, and draw the current view in this window, clearing it when changing e.g. from menu to game.~~
Edit: changed my mind, using a single window would  be difficult, because one would need to have all `_view` classes sharing their `m_window` member. For now, let's just try to only open a view's window when the view is executed, an close it when it is no longer executed.

TheoPannetier:
Done!

Issue #499, opened 2022-02-26
[Menu] Pressing key Esc during the game opens the menu


TheoPannetier:
Depends on:
- [x] #488 

Once we can switch between view, tell the game to switch to `menu_view` when `game_view` is running (via its `exec()`) and the Escape key is pressed. 
You'll want to use the `view_mode` member of `game_view` to achieve this.

No possible test for this one AFAICS 😢

TheoPannetier:
Done 🏎️

Issue #496, opened 2022-02-23
[Menu] Class `options_view` should have member of type view_mode


kausmees:

janclod:
done.

Issue #495, opened 2022-02-23
[Menu] Class menu_view should have member of type view_mode


kausmees:
Depends on:
- [x] #493 

Same as in #494, but for class `menu_view`

- [x] Make test
- [x] Solve test

```cpp
// (495) There should be a member of type view_mode
{
  menu_view mv;
  view_mode expected_next_view = view_mode::quit;
  assert(mv.what_next() == expected_next_view);
}
```

TheoPannetier:
Done!

PR #546, opened 2022-05-12
Menu task force ready to report ✌️ 

TheoPannetier:
View switching is now operational... mostly.

**Fixes:**
- #493 
- #494 
- #495 
- #496 
- #488 
- #489 
- #490 
- #446 
- #499 

This PR raises one issue, however: all windows for `game_view`, `menu_view`, `options_view`... are created at the beginning, regardless of whether the corresponding view is in use or not. This also means that we can only switch once between two views, and not back and forth yet. See #545. Solving this will require rethinking the implementation of the view classes, and I feel this PR is big enough already 😇


richelbilderbeek:
Ha, I see the problems :-)

Screenshot from 2022-05-18 16-52-10

Whatever needs to be done, the game is still in a runnable state :-)

game_view.cpp
```diff
@@ -1,11 +1,13 @@
 #include "game_view.h"
 
-#ifndef LOGIC_ONLY // that is, compiled on GitHub Actions
+#ifndef LOGIC_ONLY // so this is NOT compiled on GitHub Actions
```
richelbilderbeek, line 3: Well spotted!

options_view.cpp
```diff
@@ -1,4 +1,7 @@
 #include "options_view.h"
+#include <cassert>
```
ollyturner, line 2: This is perfect. Always use angled brackets for standard header files.

ollyturner (changes requested):
Great Work! Now some people in they really want to can leave the game. Of course they never will :) We only have one very small comment in view_mode.cpp. It was all we could find 👍 After fixing this we definitely accept!

options_view.cpp
TheoPannetier: Monkey see, monkey do is most of the time a good policy when editing code 🙈

TheoPannetier:
Ah, can't seem to find the comment in `view_mode.cpp` 😅

richelbilderbeek (approved):

view_mode.cpp
```diff
@@ -0,0 +1,12 @@
+#include "view_mode.h"
+#include "cassert"
```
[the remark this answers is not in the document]
TheoPannetier:
```suggestion
#include <cassert>
```

Issue #547, opened 2022-05-14
A `game_view` has a window size


TheoPannetier:
```cpp
// (547) A game_view has a window size
  {
    game_view gv;
    sf::Vector2f window_size = gv.get_window_size();
    assert(window_size.x == 1280);
    assert(window_size.y == 720);
  }
```

TheoPannetier:
Done in #548

PR #548, opened 2022-05-14
Windows drawn successively, not simultaneously

TheoPannetier:
**Depends on:**
- [x] #546 

**Fixes:**
- #545 
- #547 

Windows are no longer all opened at the beginning at the game*, only when one switches to the corresponding `_view`. So if one starts with game and presses [Esc], game window closes and menu window opens.

*well they are, but they are closed immediately after setup (constructor) is done.


richelbilderbeek:
It is hard to quit (but well, who wants that anyways ;-) ), maybe we'll create an Issue later.

Screenshot from 2022-05-18 18-22-17

TheoPannetier:
@richelbilderbeek what do you mean? Is it hard to click, or does the quit button not work properly?

game_view.cpp
```diff
@@ -532,7 +516,25 @@ void test_game_view()//!OCLINT tests may be many
     }
 #endif // FIX_ISSUE_246
 
-  #endif
+
+  // (545) A game window doesn't open at construction
```
richelbilderbeek, line 521: I like this style of adding the Issue number in the comment aboce the test. You'll see me copy it :-)

game_view.h
```diff
@@ -37,9 +38,25 @@ class game_view
...
   const game& get_game() const noexcept {return m_game; }
 
+  /// Get next view
+  view_mode get_next_view() const
+  {
+    return m_next_view;
+  }
+
+  // Get the size of the window
+  const sf::Vector2f get_window_size() { return m_window_size; }
```
richelbilderbeek, line 48: Yup, it is useful to be able to read the window size. 
However, it can be made more const-correct and nooexcept-correct [F.6: If your function must not throw, declare it noexcept](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-noexcept) by doing:

```c++
const sf::Vector2f& get_window_size() const noexcept { return m_window_size; }
```

Note that we return by `const&` as we do not need a copy of this data, just the original one (see also Core Guidelines).
Adding the `const` suffices in making us happy.

game_view.h
```diff
@@ -37,9 +38,25 @@ class game_view
...
+
+  // Get the window's state, for testing purposes only
+  bool is_window_open() { return m_window.isOpen(); }
```
ollyturner, line 51: This is a usefull member function indeed. However the lack of const suggests that the view if modified just when you want to see if it open. Adding the const will make this member function even better.

game_view.h
```diff
@@ -37,9 +38,25 @@ class game_view
...
+
   ///Gets a ref to m_game
   game& get_game() noexcept {return m_game; }
 
+#ifdef VIEW_SWITCH
+  // Returns what view should come next
+  const view_mode& what_next() noexcept { return m_next_view; }
```
ollyturner, line 58: Hey another one :) Add another const here to make the world a better place!

main.cpp
```diff
@@ -157,27 +161,65 @@ int main(int argc, char **argv) //!OCLINT tests may be long
...
+  options_view ov;
+  menu_view mv;
+  game_view gv(options);
+  assert(options == gv.get_options());
+
+  // Execute and switch between views
+  while (true) // I'm young and reckless
     {
-        music_off(options);
+      switch (next_view)
```
ollyturner, line 196: This is nice!

richelbilderbeek:
I have added one commit to be able to display the game in release mode; it did not compile otherwise. Now, in release mode, a warning is simply a warning, that is, not an error.

game_view.h
TheoPannetier: Thanks, I wasn't aware of F.6!
For const ref, I should know better 🤦

TheoPannetier:
```suggestion
  const sf::Vector2f& get_window_size() const noexcept { return m_window_size; }
```

game_view.h
TheoPannetier:
```suggestion
  const view_mode& what_next() const noexcept { return m_next_view; }
```

game.pro
```diff
@@ -9,20 +9,22 @@ include(game.pri)
 include(game_view.pri)
 
 # Use the C++ version that all team members can use
-CONFIG += c++11
-QMAKE_CXXFLAGS += -std=c++11
+CONFIG += c++17
+QMAKE_CXXFLAGS += -std=c++17
```
richelbilderbeek, line 14: This is scary, i.e. to expect all team members have C++17. I guess someone else put it in, so I am going to accept this.

richelbilderbeek (approved):
It was time for a window manager and you managed to add it successfully, well done!

There was one point (using C++17) that I think was from another commit, which I will ignore due to that.

Keep up the good work!

game.pro
TheoPannetier: Found the culprit https://github.com/tresinformal/game/pull/548/commits/69c8740f1e8e1aa1ada5f3816f11030453ffec46 ;)

PR #565, opened 2022-05-25
Fixed ISSUE_#457

Yangshen0325:
We fixed the ISSUE_#457: https://github.com/tresinformal/game/issues/457


game_functions.h
```diff
@@ -20,6 +20,7 @@ double get_min_y(const game &g);
 // About players
 int count_alive_players(const game& g) noexcept;
 double get_nth_player_size(const game& g, const int i);
+color get_nth_player_color(const game& g, const int i);
```
EvoLandEco, line 23: I would suggest to move this function above the `get_nth_player_size`, we plan to arrange all the free functions in lexical order, and by doing this you can reduce the amount of work of people who volunteer to sort the functions

game_functions.cpp
```diff
@@ -42,6 +42,12 @@ double get_nth_player_size(const game& g, const int i)
   return g.get_player(i).get_diameter();
 }
 
+
+color get_nth_player_color(const game& g, const int i)
```
EvoLandEco, line 46: I think it is done perfectly

EvoLandEco (approved):
Nicely done and I have no further comments about this PR

game_functions.cpp
TheoPannetier: Agreed, except on one point: the always dreaded const correctness ;)
Since this function doesn't require to change anything in either the input or output, [it should be therefore labelled](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable) `const` ;)

```suggestion
const color get_nth_player_color(const game& g, const int i)
```

game.cpp
```diff
@@ -1785,11 +1785,13 @@ void test_game() //!OCLINT tests may be many
...
   #endif // FIX_ISSUE_241
 
-  //#define FIX_ISSUE_457
+
+ #define FIX_ISSUE_457
   #ifdef FIX_ISSUE_457
+
   {
     // (457) The color of any player can be accessed easily
-    const game g;
```
TheoPannetier, line 1792: I disagree with this: changing the test is no way to solve it. Unless of course, the test is wrong, but I don't see why that would be the case here. Obtaining the colour of a player shouldn't change anything about the game or anythign inside it.

TheoPannetier (changes requested):
Nice work, but I'm not satisfied with how you've changed the test to solve it. Try again to solve it while keeping the game a `const` variable ;)

game_functions.cpp
EvoLandEco: Problem is that:
 `const color` in  .cpp, but `color` in the header file.

game_functions.h
EvoLandEco, line 23:
```suggestion
const color get_nth_player_color(const game& g, const int i);
```

game.cpp
Yangshen0325: Thanks, Theo. I added `const` before color and Tianjian taught me how this works, I learned a lot. :)

TheoPannetier: I took the freedom to put back the `const` in game in the test, it still passes 👍

PR #580, opened 2022-09-07
Use magic_enum

richelbilderbeek:
As shown in the Mob Programming :-)

### DIFF
--- a/action_type.cpp
+++ b/action_type.cpp
@@ -2,6 +2,8 @@
 #include <cassert>
 #include <sstream>
 
+#include "../magic_enum/include/magic_enum.hpp" // https://github.com/Neargye/magic_enum
+
 void test_action_type()
 {
   #ifndef NDEBUG // no tests is release
@@ -32,6 +34,8 @@ std::ostream &operator<<(std::ostream &os, const action_type t)
 
 std::string to_str(action_type this_action_type)
 {
+  return std::string(magic_enum::enum_name(this_action_type));
+  /*
   switch (this_action_type)
   {
   case action_type::turn_left:
@@ -50,4 +54,5 @@ std::string to_str(action_type this_action_type)
     assert(this_action_type == action_type::none );
   return "none";
   }
+  */
 }
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -19,6 +19,19 @@ jobs:
       - name: Install Codecov
         run: sudo pip install codecov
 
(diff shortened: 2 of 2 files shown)


action_type.cpp
```diff
@@ -2,6 +2,10 @@
 #include <cassert>
 #include <sstream>
 
+#ifdef HAS_MAGIC_ENUM // HAS_MAGIC_ENUM is defined in game.pro
```
TheoPannetier, line 5: I enjoy the politeness of not making the game fail if members have not installed magic_enum yet :)

options_view.cpp
```diff
@@ -83,14 +83,12 @@ void options_view::show()
...
-  view_mode expected_next_view = view_mode::quit;
-  assert(ov.get_next_view() == expected_next_view);
-}
-
+  // (496) There should be a member of type view_mode
+  {
+    options_view ov;
+    view_mode expected_next_view = view_mode::quit;
+    assert(ov.get_next_view() == expected_next_view);
+  }
```
TheoPannetier, line 91: Nice applicaation of the boyscout rule :+1:

action_type.cpp
```diff
@@ -32,6 +36,9 @@ std::ostream &operator<<(std::ostream &os, const action_type t)
 
 std::string to_str(const action_type this_action_type)
 {
+  #ifdef HAS_MAGIC_ENUM // HAS_MAGIC_ENUM is defined in game.pro
+  return std::string(magic_enum::enum_name(this_action_type));
+  #else
```
TheoPannetier, line 41: As said above, I do enjoy that the user gets the option to use magic_enum or not without breaking the build, but then I wonder about the usefulness of using magic_enum at all? If we are not enforcing the  use of the library, then we are just adding more lines to a code that was already functional?

Is this solution transitory? I.e., do you intend to eventually enforce use of magic_enum?

TheoPannetier (approved):
Very nice work overall, although I have a question about the future of the coexistence of magic_enum vs the old, tedious way of working with enum classes (see my comment). In any case, I'm happy to approve the changes 👍

I took the liberty of fixing the description of  #681.

action_type.cpp
richelbilderbeek: @TheoPannetier I completely agree. This solution is transitory and I do intend to eventually enforce the use of `magic_enum`. Luckily, I have my `HAS_MAGIC_ENUM` already in place :-)

Issue #366, opened 2021-09-08
A coordinate can be moved, based on direction


swom:
The following test should pass:
```c++

#ifdef FIX_ISSUE_366
  {
    coordinate c{0,0};
    double default_speed = 1;
    double direction = 0;
    c.move(direction);
    assert(c.get_x() == default_speed);
    assert(c.get_y() == 0 );

    c.reset_coords_to_zero();
    double speed = 2;
    c.move(direction, speed);
    assert(c.get_x() == speed );
    assert(c.get_y() == 0 );

    c.reset_coords_to_zero();
    double speed = 0.5;
    c.move(direction, speed);
    assert(c.get_x() == speed );
    assert(c.get_y() == 0 );

    c.reset_coords_to_zero();
    double speed = -1;
    c.move(direction, speed);
    assert(c.get_x() == speed );
    assert(c.get_y() == 0 );
  }
#endif
```

Issue #336, opened 2021-06-16
player movement souldepend on coordinates member


pohaha:
depends on:

- [ ] #321 
- [ ] #331 
The following test should pass in player.cpp
```c++

```

TheoPannetier:
Done!

PR #585, opened 2022-09-12
Player movement based on coordinates

TheoPannetier:
**Test for:**

- [x] #336 

**Fixes:**

- [x] #366
- [x] #336


coordinate.cpp
```diff
@@ -49,35 +61,41 @@ void test_coordinate() {
 
     double default_speed = 1;
     c.move(direction);
-    assert(get_x(c) == default_speed);
-    assert(get_y(c) == 0 );
+    double error_x = abs(get_x(c) - default_speed);
+    double error_y = abs(get_y(c) - 0);
+    assert(error_x < 0.00001);
+    assert(error_y < 0.00001);
```
ollyturner, line 67: We don't really need this bit. Line 69 covers all this.

coordinate.cpp
```diff
@@ -49,35 +61,41 @@ void test_coordinate() {
...
       }
 
     ///Test along y axis
     ///When direction is 0 player should
     ///only move along the y axis
-    direction = M_PI/2;
-
-    double default_speed = 1;
+    c.reset_coords_to_zero();
+    direction = M_PI / 2;
```
ollyturner, line 84: We should use a vector of directions, like line 69, to test all the cases.

coordinate.cpp
```diff
@@ -49,35 +61,41 @@ void test_coordinate() {
...
     c.move(direction);
-    assert(get_x(c) == 0);
-    assert(get_y(c) == default_speed );
+    error_x = abs(get_x(c) - 0);
+    error_y = abs(get_y(c) - default_speed);
+    assert(error_x < 0.00001);
+    assert(error_y < 0.00001);
 
-    std::vector<double> speeds{0.1,0.5,2,-0.1,-0.2,-1};
     for(const auto& speed : speeds)
```
ollyturner, line 91: We should use a vector of directions, like line 69, to test all the cases.

player.cpp
```diff
@@ -311,15 +312,17 @@ void test_player() //!OCLINT tests may be long
...
     {
         player p;
-        cordinate predicted_player_position = predict_players_movement(p);
+        coordinate predicted_player_position = predict_players_movement(p);
         p.move();
-        assert(p.get_position()==predicted_player_position);
-        assert(((predicted_player_position.m_x - get_x(p))<0.001)&&(((predicted_player_position.m_x - get_x(p))>-0.001)));
-        assert(((predicted_player_position.m_x - get_y(p))<0.001)&&(((predicted_player_position.m_x - get_y(p))>-0.001)));
+        assert(p.get_position() == predicted_player_position);
+        assert(predicted_player_position.get_x() - get_x(p) < 0.001
```
ollyturner, line 322: We should make an issue to standardise tolerance. We can see various 0s before the 1 throughout the code.

game.h
```diff
@@ -126,7 +126,6 @@ class game
 
   ///Manages collisons with walls
   player resolve_wall_collision(player p);
-
```
ollyturner, line 129: Why did you remove this line?

player.cpp
ollyturner: We should also standardise the method we use this in.

ollyturner, line 322: To be more consistent, you could check the same way as you did in coordinate.cpp on line 64.

ollyturner (changes requested):
You fixed this problem very efficiently. However, next time you could make it more consistent throughout the PR. Overall, your code was written very well!

coordinate.cpp
TheoPannetier: I have added two more less trivial directions, but I think any more would be overkill, actually.
+ you need to make a separate expectation for every direction, unlike the convenient exemples of the x=0 or y = 0 movement 😅

coordinate.cpp
TheoPannetier: removed

game.h
TheoPannetier: I don't know. Will revert.

player.cpp
EvoLandEco: To amend @ollyturner 's comment, we should also standardise the method ***of testing whether the error produced is within the tolerance***
Let's make it a separate issue.

TheoPannetier: I agree, that's a very good idea actually. I have made this an issue (#594 ), do you want to pick it :)

EvoLandEco (changes requested):
Good job! The only thing left is to make sure that you have already reverted the deletion of that line 😸

player.cpp
EvoLandEco: @TheoPannetier Sure, I will try to implement one 🚀

game.h
TheoPannetier: Oops, no I haven't! Sorry, shouldn't promise stuff before actually committing the changes 😅

game.h
```diff
@@ -126,7 +126,6 @@ class game
 
   ///Manages collisons with walls
   player resolve_wall_collision(player p);
```
TheoPannetier, line 128:
```suggestion
  player resolve_wall_collision(player p);
  
```

game.h
TheoPannetier: done now!

Issue #584, opened 2022-09-12
A view manager can read its first view as input


TheoPannetier:
```cpp
{
    view_mode first_view = view_mode::menu;
    view_manager v{first_view};
    assert(v.get_next_view() == first_view);
  }
```

TheoPannetier:
Done ✌️

Issue #583, opened 2022-09-12
A view_manager can read game options as input


TheoPannetier:
```cpp
{
    game_options options(2022, false, get_random_kam());
    view_manager v{options};
    assert(v.get_game_view().get_options() == options);
    assert(v.get_options() == options);
  }
```

TheoPannetier:
Done :v:

Issue #582, opened 2022-09-12
A view manager can run and exit


TheoPannetier:
while we cannot test all the switching between views, we can at least test that the view manager can exec and exit.

```cpp
 {
     view_manager v{view_mode::quit};
     v.exec();
  }
```

TheoPannetier:
Not fixed yet, changed test

TheoPannetier:
Fixed :v:

PR #592, opened 2022-09-15
A `view_manager` takes care of the different windows and switching betwwen them

TheoPannetier:
Fixes
- #553 
- #578 
- #582 
- #583 
- #584 

And completes Milestone 1!


game_view.cpp
```diff
@@ -408,9 +409,8 @@ void test_game_view() //!OCLINT tests may be many
     }
 
     {
-    const game_view v;
-    // game has a member function called `get_n_ticks`, which returns zero upon construction
-    assert(v.get_game().get_n_ticks() == 0);
+        game_view v;    // game has a member function called `get_n_ticks`, which returns zero upon construction
```
Yangshen0325, line 412: I think we could keep "const" there. Is there a good reason why "const" is not used?

main.cpp
```diff
@@ -161,67 +161,35 @@ int main(int argc, char **argv) //!OCLINT tests may be long
...
 
-  // Declare all views
-  options_view ov;
-  menu_view mv;
-  game_view gv(options);
-  assert(options == gv.get_options());
+  view_manager v{start_view, options};
+  v.exec();
+
+  return 0; // Game completes successfully
```
Yangshen0325, line 192: I enjoyed the simplified code there. Great!

view_manager.cpp
```diff
@@ -1,18 +1,103 @@
...
+        assert(vw.get_next_view() != next_view);
+        vw.set_next_view(next_view);
+        assert(vw.get_next_view() == next_view);
+    }
+
+    /// (582) A view manager can run and exit
     {
-        view_manager{};
+       vw.set_next_view(view_mode::quit);
+       vw.exec();
```
Yangshen0325, line 97: It seems good to re-use the same object multiple times, but it is unnecessary: It is better to have independent tests than tests that take a couple of seconds shorter. Please use a new/fresh view_manager every test. 🙂No need to optimize for speed yet 🙂 .

Yangshen0325 (changes requested):
Great job! We think you could improve even more in the way as we described in the review. If you do so, we will happily accept.

game_view.cpp
TheoPannetier: good point, fixed 👍

view_manager.cpp
TheoPannetier: gotcha, I reverted those changes to get back to independent tests.

richelbilderbeek: Beautiful! Thanks!

richelbilderbeek (approved):
Good stuff! Thanks for keeping the tests independent :-)

PR #603, opened 2022-09-28
Use int instead of short or unsigned int

### LINKED ISSUE #593: Use 'int' instead of 'unsigned int' or 'short'
In the class `scoreboard` there are those fancy datatypes called 'unsigned int' and 'short'. It is recommended to always use `int` (see Scott Meyers, 'Efficient C++' and 'C++ Coding Standards: 101 Rules, Guidelines, and Best Practices' by Sutter & Alexandrescu). 

Reasons are:

 * the one bit does not matter sizewise, 
 * the one bit is actually useful to detect errors
 * using `short` is premature optimization: we do not need to save those couple of bytes

EvoLandEco:
Fixes #593

### DIFF
--- a/scoring_board.cpp
+++ b/scoring_board.cpp
@@ -4,11 +4,11 @@
 #include <fstream>
 #include <string>
 
-scoring_board::scoring_board(unsigned int score_player1,
-                             unsigned int score_player2 ,
-                             unsigned int score_player3,
-                             unsigned int timer,
-                             short winner)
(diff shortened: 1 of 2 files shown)


richelbilderbeek (approved):
Excellent! Thanks for [not optimizing prematurely](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rper-reason) :-)

PR #610, opened 2022-09-28
Add install_windows.md for Windows users to install correct version of Qt and MinGw

### LINKED ISSUE #605: Update install.md for Windows
We need to the update the installation guide for windows because it is already outdated. The required versions of QT as well as MinGW now belong to the archive, we should manually refresh the list to be able to see them.

EvoLandEco:
Fixes  #605


richelbilderbeek (approved):
This is a good as it gets. Keep up the good work!

Issue #628, opened 2022-10-02
A respawn color can be picked randomly


TheoPannetier:
### Context

To respawn a player with a new colour (#612), we need to be able to sample a color at random between red, green, blue. 

### Test
- [x] Test is ready and on develop 

```cpp
// (628) A random color between red, blue, green can be sampled
  {
    const color r = create_red_color();
    const color g = create_green_color();
    const color b = create_blue_color();

    const int a_seed{42};
    std::mt19937 rng1(a_seed);
    const color a_color = get_random_respawn_color(rng1);
    assert(a_color == r || a_color == g || a_color == b);

    const int another_seed{43}; /// please change seed if draws same color as 42
    std::mt19937 rng2(another_seed);
    const color another_color = get_random_respawn_color(rng2);
    assert(another_color == r || another_color == g || another_color == b);
    assert(a_color != another_color);

    const int same_seed = a_seed;
    std::mt19937 rng3(same_seed);
    const color same_color = get_random_respawn_color(rng3);
    assert(same_color == a_color);
  }
```

richelbilderbeek:
@TheoPannetier: I enjoy the enthusiasm of this big test! Using for-loops in tests, however, usually is a code smell. In your suggested case, I feel the test is more focussed on the C++ RNG, which we **should** assume is correct.

 I think the focus of the test is different: two randomly-drawn colors should differ.

Following that, the tests it can be simplied quite a bit. Here I'll suggest how.

One approach is to use `std::rand` in the back. You should never use `std::rand` in important code, but AFAICS, we use those random colors only in debugging. That would be my definition of not important.

```c++
// Check if the color is actually random
{
   const auto c = get_random_color();
   const auto d = get_random_color();
   assert(c != d); // Chance 1 in 256^3 this fails
}
```

I think this a great test with using a RNG engine:

```c++
// Check if the color is actually random
{
  const int seed{42}
   std::mt19937 rng(seed);
   const auto c = get_random_color(rng); //Will modify RNG
   const auto d = get_random_color(rng); //Will modify RNG
   assert(c != d); // Chance 1 in 256^3 this fails, pick a different seed if needed
}
// Check if the color is actually random yet repeatable
{
   const int seed{42}
   std::mt19937 rng1(seed);
   std::mt19937 rng2(seed);
   const auto c = get_random_color(rng1); //Will modify RNG
   const auto d = get_random_color(rng2); //Will modify RNG
   assert(c == d);
}
```

[...]

Issue #626, opened 2022-10-01
A player can be revived


TheoPannetier:
### Context

After some time, a player that has been killed should be allowed to revive and join the game again.
A first step is to make a function that can revive a player.

### Test
- [ ] Test is ready and on develop 

```cpp
#define FIX_ISSUE_626
{
  
}
#endif // FIX_ISSUE_626
```

**Tip:** look for the definition of `kill_player()`

MahshidGhods:
I'm a member now.

Issue #625, opened 2022-10-01
A player that is out cannot collide with other players


TheoPannetier:
### Context

While a player is out, as a result of being killed, it should not be able to run in collisions with other players, instead passing through them as if it didn't exist.

### Test
- [x] Test is ready and on develop 

```cpp
#define FIX_ISSUE_625
{
      // (625) A player that is out cannot collide with other players
    game g;
    const coordinate c_p2 = g.get_player(1).get_position();
    g.get_player(0).place_to_position(c_p2);
    assert(are_colliding(g.get_player(0), g.get_player(1)));
    g.kill_player(0);
    assert(!are_colliding(g.get_player(0), g.get_player(1)));
}
#endif // FIX_ISSUE_XXX
```

Issue #624, opened 2022-09-29
A player that is out is immune to stun rockets


TheoPannetier:
### Context

A player that has been killed should be immune to stun rockets, and should not stop their trajectory!

### Test
- [ ] Test is ready and on develop 

```cpp
// (624) A player that is out cannot be stunned
    // and does not absorb stun rockets
    game g;
    g.kill_player(0);

    // Player 2 shoots, rocket goes on player 1
    g.do_action(1, action_type::shoot_stun_rocket);
    g.tick();
    const coordinate c_p1 = g.get_player(0).get_position();
    g.get_projectiles().back().place(c_p1);

    g.tick();
    assert(!is_stunned(g.get_player(0)));
    assert(count_n_projectiles(g) == 1);
```

PR #629, opened 2022-10-03
Tests for Milestone 3: killing and reviving player

### LINKED ISSUE #627: A player's state is changed by specific functions, rather than a setter
### Context

So far, we use `player::set_state()` to change a player's states, e.g. to kill, revive or stun it. This is ok, but a more robust approach would be to change state only with a member function that takes care of the specific action.

This is particularly important if the logic of the game requires other things to happen when a state is changed, e.g. changing colour when a player dies or trigger a timer when the player is stun.

### Test
- [ ] Test is ready and on develop 

```cpp
{
  
}
```

TheoPannetier:
I have made a bunch of tests from the chat we had with @ollyturner last week, to translate the ideas from the game design team into actual issues. There is a bunch more to do, but this is good to go on `develop` so that juniors can start solving the issues :)

**Tests for:**

- [ ] #606 
- [ ] #607 
- [ ] #608 
- [ ] #609 
- [ ] #611 
- [ ] #621 
- [ ] #622 
- [ ] #623 
- [ ] #624 
- [ ] #625 
- [ ] #626 
- [ ] #627 
- [ ] #628 

I also went ahead and already fixed #627 😇

### DIFF
--- a/color.cpp
+++ b/color.cpp
@@ -215,10 +215,9 @@ void test_color()
   }
 #endif // FIX_ISSUE_460
 
-#ifdef FIX_ISSUE_628
+  #ifdef FIX_ISSUE_628
   // (628) A random color between red, blue, green can be sampled
   {
-    color c;
     int n_red = 0;
     int n_green = 0;
     int n_blue = 0;
@@ -230,7 +229,7 @@ void test_color()
     std::mt19937 rng(1);
     for (int i = 0; i < 100; ++i)
       {
-        c = get_random_rgb(rng);
+        const color c = get_random_rgb(rng);
         if (c == r) { ++n_red; }
         else if (c == g) { ++n_green; }
         else if (c == b) { ++n_blue; }
--- a/color.cpp
+++ b/color.cpp
@@ -4,6 +4,7 @@
 #include <cstdlib>
 #include <sstream>
 #include <iostream>
+#include <random>
 #include <SFML/Graphics.hpp>
 
 color::color(const int r, const int g, const int b, const int a)
@@ -214,6 +215,33 @@ void test_color()
   }
 #endif // FIX_ISSUE_460
 
+#ifdef FIX_ISSUE_628
+  // (628) A random color between red, blue, green can be sampled
+  {
+    color c;
+    int n_red = 0;
+    int n_green = 0;
+    int n_blue = 0;
+
+    const color r = create_red_color();
+    const color g = create_green_color();
+    const color b = create_blue_color();
+
+    std::mt19937 rng(1);
+    for (int i = 0; i < 100; ++i)
+      {
+        c = get_random_rgb(rng);
+        if (c == r) { ++n_red; }
+        else if (c == g) { ++n_green; }
+        else if (c == b) { ++n_blue; }
+        else { throw("get_random_rgb should not return any other color than r, g, b"); }
+      }
+
+    assert(n_red > 0 && n_red < 100);
+    assert(n_green > 0 && n_green < 100);
+    assert(n_blue > 0 && n_blue < 100);
+  }
+#endif // FIX_ISSUE_628
 
 
 #endif // NDEBUG
--- a/game.cpp
+++ b/game.cpp
@@ -335,7 +335,7 @@ void game::kill_player(const int index)
 {
   assert(index >= 0);
   assert(index < static_cast<int>(m_player.size()));
-  get_player(index).set_state(player_state::out);
+  get_player(index).die();
 }
 
 void game::resolve_wall_collisions()
@@ -1349,6 +1349,21 @@ void test_game() //!OCLINT tests may be many
   }
 #endif
 
+#ifdef FIX_ISSUE_611
+  {
+    // (611) A player that is out revives after some time
+    game g;
+    g.kill_player(0);
+    const int revive_time = 100;
+    for (int i = 0; i < revive_time; ++i)
+      {
+        assert(is_out(g.get_player(0)));
+        g.tick();
+      }
+    assert(!is_out(g.get_player(0)));
+  }
+#endif
+
 #ifdef FIX_ISSUE_606
   {
     // (606) When a player goes under some size, it is out
@@ -1846,8 +1861,24 @@ void test_game() //!OCLINT tests may be many
     game g;
     assert(get_nth_player_color(g, 0) == create_red_color());
     g.kill_player(0);
-    g.tick();
     assert(get_nth_player_color(g, 0) == create_white_color());
+  }
+#endif
+
+#define FIX_ISSUE_612
+#ifdef FIX_ISSUE_612
+  {
+    // (612) A player that revives gets a new colour
+    assert(1 == 2);
+    game g;
+    g.kill_player(0);
+    const int revive_time = 100;
+    for (int i = 0; i < revive_time; ++i)
+      {
+        assert(is_out(g.get_player(0)));
+        g.tick();
+      }
+    assert(!is_out(g.get_player(0)));
   }
 #endif
 
--- a/game.cpp
+++ b/game.cpp
@@ -1335,55 +1335,55 @@ void test_game() //!OCLINT tests may be many
     g.kill_player(0);
 
     assert(num_of_players_begin == g.get_v_player().size());
-    assert(is_out(g.get_player(0)));
+    assert(is_dead(g.get_player(0)));
   }
 
-#ifdef FIX_ISSUE_606
+#ifdef FIX_ISSUE_626
   {
-    // (626) A player that is out can be revived
+    // (626) A player that is dead can be revived
     game g;
     g.kill_player(0);
-    assert(is_out(g.get_player(0)));
+    assert(is_dead(g.get_player(0)));
     g.revive_player(0);
-    assert(!is_out(g.get_player(0)));
+    assert(!is_dead(g.get_player(0)));
   }
 #endif
 
 #ifdef FIX_ISSUE_611
   {
-    // (611) A player that is out revives after some time
+    // (611) A player that is dead revives after some time
     game g;
     g.kill_player(0);
     const int revive_time = 100;
     for (int i = 0; i < revive_time; ++i)
       {
-        assert(is_out(g.get_player(0)));
+        assert(is_dead(g.get_player(0)));
         g.tick();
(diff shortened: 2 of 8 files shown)


action_type.cpp
```diff
@@ -30,7 +30,7 @@ std::ostream &operator<<(std::ostream &os, const action_type t)
   return os;
 }
 
-std::string to_str(action_type this_action_type)
+std::string to_str(const action_type this_action_type)
```
richelbilderbeek, line 33: :+1:

color.cpp
```diff
@@ -214,6 +215,33 @@ void test_color()
...
+    int n_blue = 0;
+
+    const color r = create_red_color();
+    const color g = create_green_color();
+    const color b = create_blue_color();
+
+    std::mt19937 rng(1);
+    for (int i = 0; i < 100; ++i)
+      {
+        c = get_random_rgb(rng);
```
richelbilderbeek, line 233: One could write `const color c = ...` here. It is a detail, but let's show the newbies our best, shall we :-) ?

color.cpp
```diff
@@ -214,6 +215,33 @@ void test_color()
...
+        if (c == r) { ++n_red; }
+        else if (c == g) { ++n_green; }
+        else if (c == b) { ++n_blue; }
+        else { throw("get_random_rgb should not return any other color than r, g, b"); }
+      }
+
+    assert(n_red > 0 && n_red < 100);
```
richelbilderbeek, line 240: I see the idea here; it would be weird if `n_red` would be more than 100. However, that would mean we are testing the test itself. I feel that is overkill.

Hence I propose to use `assert(n_red > 0);`, as this resembles test best ('At least once, we should get red'). To add more weight behind this argument, adding the second part results in a needless fail when a user increases the number of for loop iterations to e.g. 1000.

game.cpp
```diff
@@ -223,7 +223,7 @@ void game::projectile_collision()
 
               // if the projectile is a stun rocket: stun the player
               if(this-> m_projectiles[i].get_type() == projectile_type::stun_rocket)  {
-                  this-> m_player[j].set_state(player_state::stunned);
+                  this-> m_player[j].stun();
```
richelbilderbeek, line 226: I love this expressiveness!

game.cpp
```diff
@@ -336,7 +335,7 @@ void game::kill_player(const int index)
 {
   assert(index >= 0);
   assert(index < static_cast<int>(m_player.size()));
-  get_player(index).set_state(player_state::dead);
+  get_player(index).die();
```
richelbilderbeek, line 338: Beautiful!

game.cpp
```diff
@@ -1351,18 +1326,72 @@ void test_game() //!OCLINT tests may be many
...
-    assert(is_dead(g.get_player(0)));
+    assert(is_out(g.get_player(0)));
+  }
+
+#ifdef FIX_ISSUE_606
+  {
+    // (626) A player that is out can be revived
+    game g;
+    g.kill_player(0);
+    assert(is_out(g.get_player(0)));
```
richelbilderbeek, line 1346: I am unsure why `is_out` is used over `is_dead`. I assume there is some smart reasoning behind this :-)

game.cpp
```diff
@@ -1351,18 +1326,72 @@ void test_game() //!OCLINT tests may be many
...
+    g.revive_player(0);
```
richelbilderbeek, line 1347: Excellent function name choice here, `revive` expresses this well!

game.cpp
```diff
@@ -1351,18 +1326,72 @@ void test_game() //!OCLINT tests may be many
...
+    assert(!is_out(g.get_player(0)));
+  }
+#endif
+
+#ifdef FIX_ISSUE_611
+  {
+    // (611) A player that is out revives after some time
+    game g;
+    g.kill_player(0);
+    const int revive_time = 100;
```
richelbilderbeek, line 1357: Where does `revive_time` come from? Either add `// just a high number` or use a function such as `get_default_revive_time`.

game.cpp
```diff
@@ -1351,18 +1326,72 @@ void test_game() //!OCLINT tests may be many
...
+      }
+    assert(!is_out(g.get_player(0)));
+  }
+#endif
+
+#ifdef FIX_ISSUE_606
+  {
+    // (606) When a player goes under some size, it is out
+    game g;
+    player p = g.get_player(0);
```
richelbilderbeek, line 1371: I see the idea here and I agree with that.

However, there is a problem here: the `player p` is a **copy** of the first player in the game. This means that `g.tick();` will not operate on `p`. I predict that, due to this, this test does not even fail.

Solutions are to use either `player& p = g.get_player(0);` or use `g.get_player(0)`/`get_first_player(g)` every time.

game.cpp
```diff
@@ -1351,18 +1326,72 @@ void test_game() //!OCLINT tests may be many
...
+        p.shrink();
+        g.tick();
+      }
+    assert(is_out(p));
   }
+#endif
+
+  #ifdef FIX_ISSUE_625
+    {
+    // (625) A player that is out cannot collide with other players
```
richelbilderbeek, line 1386: Fun test, I like this!

game.cpp
```diff
@@ -1375,7 +1404,19 @@ void test_game() //!OCLINT tests may be many
     assert(!has_any_player_food_collision(g));
   }
 
-  // #392: When a player gets within the radius of food it eats it
+#ifdef FIX_ISSUE_621
+  // (621) A player that is out does not eat food
+  {
+    game g;
+    put_player_on_food(g.get_player(0), g.get_food()[0]);
```
richelbilderbeek, line 1411: This works fine! Maybe this can be made more expressive by using a simple function called `put_first_player_on_first_food`, but you decide here :-)

game.cpp
```diff
@@ -1745,13 +1783,65 @@ void test_game() //!OCLINT tests may be many
...
-    //THIS LINE DOESN't WORK
     assert(count_n_projectiles(g) == 0);
 
     // Player 2 is now stunned yet
     assert(is_stunned(g.get_v_player()[1]));
   }
 
+#ifdef FIX_ISSUE_622
+  {
+    // (622) A player that is out cannot shoot rockets
```
richelbilderbeek, line 1794: Again, I enjoy the looking ahead of things that may happen in the game!

player.cpp
```diff
@@ -178,24 +178,42 @@ bool is_red(const player & p) noexcept
             ;
 }
 
+// The player can be stunned
+void player::stun()
```
richelbilderbeek, line 182: Yay to expressiveness!

player.cpp
```diff
@@ -586,6 +604,56 @@ void test_player() //!OCLINT tests may be long
...
+      p.die();
+      p.revive();
+      c = p.get_color();
+      if (c == r) { ++n_red; }
+      else if (c == g) { ++n_green; }
+      else if (c == b) { ++n_blue; }
+      else { throw("A player should not revive with any other color than r, g, b"); }
+    }
+
+  assert(n_red > 0 && n_red < 100);
```
richelbilderbeek, line 651: Same here, no need for `n_red < 100` :-)

richelbilderbeek (changes requested):
Such a fine commit! I especially enjoy the looking ahead, i.e. the problems that may arise in the game in the future, such as being ~~dead~~ out and that player trying to shoot.

I did have some minor changes here, as I predict the authors would enjoy me to be picky. If those are fixed or discussed back, I'd happily accept. Keep up the good work!

color.cpp
TheoPannetier: Isn't it going to be an issue when I change the value of `c` in the for loop? I thought `const` was only for objects that don't change value after being declared

color.cpp
TheoPannetier: I meant to check that not all iterations (that is, less than 100) result in red. But indeed, this is redundant with the next lines and adds needless complexity!

color.cpp
```diff
@@ -214,6 +215,33 @@ void test_color()
...
+    assert(n_green > 0 && n_green < 100);
+    assert(n_blue > 0 && n_blue < 100);
```
TheoPannetier, line 242:
```suggestion
```

color.cpp
TheoPannetier, line 240:
```suggestion
    assert(n_red > 0 && n_green > 0 && n_blue > 0 );

```

color.cpp
richelbilderbeek: Yes, your believes are correct! The compiler should not complain when making this `const`. Or did I miss something???

game.cpp
TheoPannetier: Yeah I thought "out" is more appropriate since "dead" would imply the player is eliminated. Byt reading through the game logic it doesn't sound so; I think I'm goin to revert back to "dead"

game.cpp
TheoPannetier, line 1357:
```suggestion
    const int revive_time = 100; // arbitrary; choose a better number when solving this test :) 
```

TheoPannetier: Agreed; I think I mention in the issue that the nb is arbitrary, but not in code yet

player.cpp
```diff
@@ -586,6 +604,56 @@ void test_player() //!OCLINT tests may be long
...
+  assert(n_green > 0 && n_green < 100);
+  assert(n_blue > 0 && n_blue < 100);
```
TheoPannetier, line 653:
```suggestion
```

player.cpp
TheoPannetier, line 651:
```suggestion
  assert(n_red > 0 && n_green > 0 && n_blue > 0);
```

game.cpp
TheoPannetier: Oops, indeed! I should know that. Conveniently, `get_player` returns a reference to the player 😏

game.cpp
TheoPannetier: Nah, I don't feel its is useful to add a function just for that one test :)

color.cpp
TheoPannetier: Ok, I got this, I'll just remove the declaration above and instead declare `c` inside the for loop

richelbilderbeek (approved):
Well done! Thanks for the exemplary work!

PR #634, opened 2022-10-05
Progress #631

### LINKED ISSUE #631: An options_view allows a user to modify game_options
### Context

The options screen (called `options_view`) is a dialog in which the user can specify his/her ideal preferences for the game.
These preferences/settings are encapsulated in `game_options`.
It would make sense that the `options_view` allows one to modify all `game_options`.

In this case, provide for:

 * a `options_view` takes a `game_options` at construction, which will be displayed on screen
 * a `options_view` has an internal `game_options` it works on
 * a `options_view` has an internal `game_options` that can be requested at, for example, when closing the dialog

### Test

 * [x] On `develop`? Not yet

In `options_view.cpp`:

```cpp
  #ifdef FIX_ISSUE_631
  // (631) An options_view allows a user to modify game_options
  {
    const game_options options;
    const options_view view(options);
    const game_options options_again(view.get_options());
    assert(options == options_again);
  }
  #endif // FIX_ISSUE_631
```

richelbilderbeek:
This is partial progress to fix #631, as shown in the mob programming.

### DIFF
--- a/doc/cpp_core_guideline_of_the_month.md
+++ b/doc/cpp_core_guideline_of_the_month.md
@@ -1,0 +1,5 @@
+# C++ Core Guideline of the month
+
+Month        |Notes
+-------------|-----------------------------------------------
+Setember 2022|https://github.com/tresinformal/game/issues/586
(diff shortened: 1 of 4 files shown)


OggyOggelito (approved):
Approve and well done Richel! Good solution

richelbilderbeek:
I will merge as (me and @OggyOggelito) forgot to do so yesterday.

PR #647, opened 2022-11-02
Implement are_equal

### LINKED ISSUE #594: Equality between `double`s can be checked, with some tolerance
`double`s (i.e., numbers with decimal values) can be checked for equality with `x == y`, but this will only return `TRUE` if both numbers are EXACTLY equal. 
Because of rounding error or other forms of imprecision, often in tests we are forced to use something like `abs(x - y) < 0.0001`, that is the difference is smaller than some level of tolerance, which we tend to define arbitrarily and for every test.

Instead, it would be convenient for tests to be able to call a function that checks for this, with a preset level of tolerance.

1. Create a test for such a function `are_equal()` that would return TRUE if the difference between two doubles is smaller than some tolerance
2. Create this function :)
3. Replace all occurrences of tests of equality between doubles with this function.

You can make the level of tolerance hard-coded in the function definition, an argument of the function, or both, whatever you think is more advisable. You're boss!

EvoLandEco:
Fixes #594


install_windows.md
```diff
@@ -36,7 +36,7 @@ Here, you will need to check two boxes:
  * 2. Now we can see Qt 5.14.2, we tick it and expand it so that we could also see the correct version of MinGw (7.3.0).
 ![image](https://user-images.githubusercontent.com/57348932/192825152-980e947e-3cbc-46f7-a0b1-c0ac28f2a0a6.png)
 
-## 2. Download and install SFML ([here](https://www.sfml-dev.org/files/SFML-2.5.1-macOS-clang.tar.gz))
```
EvoLandEco, line 39: I have no idea where this change comes from, might be somewhere in develop

utils.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include <cassert>
+#include <cmath>
+
+bool are_equal(const double a, const double b, const double tolerance = 0.00001);
+
+void test_utils();
+
+#endif
+
```
EvoLandEco, line 12:
```suggestion
```

install_windows.md
richelbilderbeek: No idea either. It;'s OK :-)

utils.cpp
```diff
@@ -0,0 +1,21 @@
...
+    return std::abs(a - b) < tolerance;
+}
+
+void test_utils()
+{
+#ifndef NDEBUG // no tests in release
+
+  // (594) Equality between doubles can be checked, with some tolerance
+  {
+    assert(are_equal(0.0, 0.0));
```
richelbilderbeek, line 14: Excellent tests!

richelbilderbeek (approved):
Good stuff! Keep up the good work!

PR #648, opened 2022-11-02
[MVG] Remove key bindings of shoot and shoot stun rocket

### LINKED ISSUE #644: [MVG] Remove key bindings for shooting rocket and stun rocket
To make a minimal viable game (MVG), we have to implement and also remove some functionalities.

We will remove rocket and stun rocket from the game by commenting out the key bindings. We think that this approach would be easier and has less impact on the whole program.

EvoLandEco:
fixes #644 @ollyturner


install_mac.md
```diff
@@ -26,10 +26,19 @@ Here, you will need to check two boxes:
...
+     type `cd /Users/[username]/SFML`, where `[username]` is your
+     username, resulting in, for example, `cd /Users/richel/SFML`.
+     This should give no message. If the message is 'Folder not found', 
  * Copy the content of `SFML/Frameworks` folder into `/Library/Frameworks`.
    You will need root rights for this
-
+   * Test that it works: 
+     In a Terminal, 
+     type `cd /Library/Frameworks && ls`.
+     You should see folders that are named something like `sfml-framework`.
```
EvoLandEco, line 41: Not my commit, might be from develop

TheoPannetier: That's odd indeed, but it doesn't appear to be from develop, otherwise it wouldn't show up here.
In any case that looks like a strict upgrade :)

key_action_map.cpp
```diff
@@ -455,8 +455,9 @@ void test_key_action_map()//!OCLINT tests can be many
     assert(kam.to_key(action_type::turn_right) == sf::Keyboard::D);
     assert(kam.to_key(action_type::accelerate_forward) == sf::Keyboard::W);
     assert(kam.to_key(action_type::accelerate_backward) == sf::Keyboard::S);
-    assert(kam.to_key(action_type::shoot) == sf::Keyboard::Q);
-    assert(kam.to_key(action_type::shoot_stun_rocket) == sf::Keyboard::E);
+    // Note that this line is commented out due to the need of minimal viable game, ref to (644)
+    //assert(kam.to_key(action_type::none) == sf::Keyboard::Q);
```
TheoPannetier, line 459: I don't understand why you turn the link to Q off, but not E?
Also, commenting out code is a sloppy policy, version control has your back so be bold and delete entirely if you don't need it 😎

OggyOggelito (commented):
I approve all the changes. But key binding Q should be activated?

TheoPannetier (commented):
I approve of the trade off to link rocket keys to <<nothing>> until we get a minimal version, but there is still a part of the test I don't understand, could you help me clarify what you want to do there?

install_mac.md
EvoLandEco: Really odd because I first merged develop into mine and then added new things

key_action_map.cpp
EvoLandEco: Will delete those lines asap, but I will be unavailable until 17th.

I think I turned off both Q and E, if I understand the minimal viable game goal correctly

TheoPannetier: Yes, no rush :)
I can also do it on your behalf, let me suggest a change.
But I'm still unsure what you want to do exactly. If it's just deactivating the rockets, you can keep keys Q and E, mapped to `action_type::none`. That should mean that pressing those keys triggers... nothing! right?

TheoPannetier, line 459:
```suggestion
    assert(kam.to_key(action_type::none) == sf::Keyboard::Q);
```

EvoLandEco: Hey, I tried again and found out why I commented this out, the reason is if not then the test will fail, I don't quite get it but this is a necessary step, and I kept it because I haven't figured out the importance of this test, so I just kept 😸

EvoLandEco: Yes, and I think this will be the neatest way to disable these two keys, without breaking to many things

TheoPannetier: I've figured out why :)
In this test, we are testing a function (`key_action_map::to_key`) that given a type of action, returns the key that is mapped to it.
Here, we mapped two keys to the same action type (`action_type::none`), so I suppose the second one (E) overrides the first (Q) when called. E is returned both times.

TheoPannetier: I feel there is no point of testing which key "nothing" corresponds to, so I suggest simply removing both lines for now, and resurrect them (mapped to stun rocket and rocket) after we are done with the MVG version.

EvoLandEco: removed these tests

TheoPannetier (approved):
Hey, idk if you appreciate another unsolicited review, but I'm happy with the changes  and would approve the merge :+1:
I won't merge bc I guess you want to hear @richelbilderbeek 's opinion on `action_type::none`, but otherwise, feel free to merge yourself!

EvoLandEco:
> Hey, idk if you appreciate another unsolicited review, but I'm happy with the changes and would approve the merge 👍 I won't merge bc I guess you want to hear @richelbilderbeek 's opinion on `action_type::none`, but otherwise, feel free to merge yourself!

@TheoPannetier @richelbilderbeek Thanks for the review, I think we can always do things step-by-step:

1. Change the key bindings to `none`
2. Change `none` to `idle`
3. Ultimately, we could adopt `std::optional`

For now I believe the solution in this PR is at least straight and will not break the other part of the code, let's merge this PR first and keep track of the rest in the issues 😸

Issue #599, opened 2022-09-23
`about` contains a link to the the GitHub repo


TheoPannetier:
It would fun to offer players the possibility to contribute to the game, or just see the code via a button that direct to this repo.

For this, `about` should contain a link to the repo, to be passed to the corresponding button in `about_view`

Issue #598, opened 2022-09-23
`about` contains a link to join the Discord server


TheoPannetier:
I feel the About screen should offer interested players to join and contribute. This could be done via a "Join" button that links to an invite to the Discord server.

For this, we need `about` to get an invite link, to be passed to `about_view` later on.

```cpp
{
    about ab;
    std::vector<std::string> link_to_discord = ab.get_link_discord();
 }
```

Luxwor:
its proving really dificult

OggyOggelito:
There must be a way to dodge

Issue #596, opened 2022-09-23
`about` keeps track of the game version


TheoPannetier:
Users should be able to read which release of the game they are playing. The game should keep this info in `about`, to be displayed on screen via `about_view`.

The current version can be found in the [release ](https://github.com/tresinformal/game/releases)tab

```cpp
  {
    about ab;
    std::vector<std::string> version = ab.get_version();
  }
```

Issue #433, opened 2021-11-17
Call the graphical display `about_view`


swom:

TheoPannetier:
done!

TheoPannetier:
Not done, my bad!

TheoPannetier:
Done for good this time

PR #649, opened 2022-11-09
Progress on implementing the About screen

TheoPannetier:
Fixed:

- [x] #433 

Tests for:

- [ ] #596 
- [ ] #599 
- [ ] #598 

I believe some have been addressed already by eager juniors, so the tests arrive a bit late 😅


OggyOggelito (commented):
I approve these changes! Good work Theo

Issue #640, opened 2022-10-12
Add title to options screen


richelbilderbeek:
In big letters

PR #651, opened 2022-11-16
Add test for #640

richelbilderbeek:


richelbilderbeek:
I am going to bypass review, as I feel we have too few tests at the moment. I may regret that in the future, but let's observe that :-)

Issue #643, opened 2022-10-19
Players' stats and event logs


EvoLandEco:
Player class may or may not need a score member.

If the scores are calculated based on event logs and player stats, then the scoring board itself can do the job. We may have the scoring board to keep the K/D/As and other stats for the players and synthesize these with the event logs (which may be implemented as a new class) to get the final scores and set the final winner(s).

But one may also think that in the above case we give too much power to the scoring board. An alternative approach is to let the players keep their own stats and records and report them to the scoring board when enquired.

To sum up, we need to decide where to keep the event logs and players' stats.

richelbilderbeek:
Great idea! Write one or more tests to put your ideas into code :+1:

EvoLandEco:
@richelbilderbeek I am looking for a proper way to design the event logs, e.g., how to store the entries:

A player eats a food, we need an entry with `(player_id, food_type, amount, n_ticks)`,
A player hits another player, we need `(player_id, another_player_id, n_ticks)`
A player collects an item, we need `(player_id, item_type, amount, n_ticks)`

And maybe much more different combinations in the future, `std::variant` seems legit here, but I don't think we are using c++ 17 standard? Could you please give some advice on this?

richelbilderbeek:
@EvoLandEco, my advice will be: (1) write a first simple test. (2) expand from there, while keeping the test.

I would enjoy seeing the reasoning behind using `std::variant`, a log message seems like a `std::string` to me :-)

We do use C++17, you can see that in `game.pro`.

I hope that helps. Again, start writing simple tests.

Note that this Issue is actually two: (1) a player has stats, (2) the game has an event log. About (2): Do we really need a log? If yes, write a simple test.

Again, I hope that helps. Again, start writing simple tests.

EvoLandEco:
> @richelbilderbeek Sure, I'll write a few tests today to illustrate the ideas, and I think I just thought too much, and it would be the best to just keep things as simple as possible, such as, parsing a simple` std::string`

And yes, event log may not be really needed as we could just let things happen without recording it in a separate place

richelbilderbeek:
@EvoLandEco yup, writing tests prevents one from overthinking :-) . Welcome to The Good Life :+1:

PR #653, opened 2022-11-23
Towards a (more) functional scoring board

EvoLandEco:
dependent #643 

In this PR I will try to work towards a functional scoring board, and I think a `event_type` class would be a major step, please share your ideas on this, especially if you think this is not a good way


EvoLandEco:
This PR is done, will do the rest later, to not make it too big for a PR

richelbilderbeek (approved):
Good stuff! Although we won't use the scoreboard until after the minimal viable version, it does not hurt to have that tested class around :-)

PR #654, opened 2022-11-23
Fixed Issue #628

ollyturner:
We use mt19937 to create a (pseudo!) random number.
We use time as a seed where possible.
We follow naming suggestions from @TheoPannetier
Some old stuff, we don't fully understand. This is from May!


color.cpp
```diff
@@ -56,6 +55,15 @@ color create_black_color()
   return color(0, 0, 0, 255);
 }
 
+color get_random_respawn_color(std::mt19937& ran_num_gen) noexcept
+{
+    int r = 1 + ran_num_gen()/((ran_num_gen.max() + 1u)/255);
+    int g = 1 + ran_num_gen()/((ran_num_gen.max() + 1u)/255);
+    int b = 1 + ran_num_gen()/((ran_num_gen.max() + 1u)/255);
```
richelbilderbeek, line 62: I'd be impressed if this actually works. Note that the range is *never* tested. I'll happily write an Issue to write those :-)

color.h
```diff
@@ -46,6 +47,9 @@ int get_opaqueness(const color &c) noexcept;
 /// Get the redness
 int get_redness(const color &c) noexcept;
 
+/// Get random color using random number generator
+color get_random_respawn_color(std::mt19937& ran_num_gen) noexcept;
```
richelbilderbeek, line 51: I enjoy the name! It felt weird at first, but it is indeed the color a player gets when spawning.

coordinate.cpp
```diff
@@ -31,6 +32,12 @@ bool coordinate::operator!=(coordinate in_coord) {
   return not((m_x == in_coord.m_x) && (m_y == in_coord.m_y));
 }
 
+std::ostream& operator << (std::ostream &out, const coordinate &cor)
+{
+    return out << cor.get_x() << " " << cor.get_y() << std::endl;
```
richelbilderbeek, line 37: I didn't know you could do this in a one-liner.

Note that the `\n` is a bad idea, as it forces the user of `operator<<` to use a newline and -sometimes- removing it. I'll create an Issue to fix that :-)

richelbilderbeek (approved):
Let's get this merged! Not the cleanest PR, but a step forward anyways. Try to merge more often :-)

PR #675, opened 2022-12-14
Can pick a random spawn color, fix #628

### LINKED ISSUE #628: A respawn color can be picked randomly

richelbilderbeek:
After some discussion, we landed at these tests, with these solutions.

We made the tests a bit cleaner, i.e. with `{}` instead of `=`, due to the C++ Core Guideline of this Month :-)


TheoPannetier (approved):
Excellent work, I like your creative solution and I approve of the stricter assignment :)

Issue #676, opened 2022-12-14
Add operator== for player to player


richelbilderbeek:
### Context

Currently, we cannot compare `player` for equality.

### Test

- [ ] Test is ready and on develop 

```cpp
#define FIX_ISSUE_676
// (676) Can compare players for equality
{
  const player a;
  const player b;
  assert(a == b);
}
#endif // FIX_ISSUE_676
```

PR #677, opened 2022-12-14
Fix #662 and #676

### LINKED ISSUE #662: Refactor: game::get_v_player -> game::get_players
Currently, the `game` has a member function called `get_v_player` that returns the players, where `v` stands for `vector`.
However, [NL.5: Avoid encoding type information in names](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-name-type), hence this should be simply renamed to `get_players`. 

To transition, both function names will be temporarily supported:

```c++
  #ifdef FIX_ISSUE_662
  {
    const game g;
    const auto players{g.get_players()};
    const auto v_player{g.get_v_player()};
    assert(players == v_player);
  }
  #endif // FIX_ISSUE_662
```

If this is done, go to #663

richelbilderbeek:
A simple transition to use less type information in the names


janclod:
Everything seems good :) well done!

Issue #671, opened 2022-12-14
Remove the projectiles and stunning from the game for now


TheoPannetier:
We don't need rockets, shooting, or stunning in the minimum viable game. We should remove it for now, and bring it back later if we want some time after the release of v1.0.

TheoPannetier:
Done!

Issue #670, opened 2022-12-14
Remove the scoreboard for the minimal viable game


TheoPannetier:

TheoPannetier:
Done!

PR #678, opened 2022-12-14
Removing scoring board, shooting, stunning and projectiles

TheoPannetier:
We don't need any of the above in the minimum viable game, and because it is indeed minimum, it should go :)

**Fix:**
- #671 
- #670


richelbilderbeek (approved):
It's hard to criticize the deletion of code :-) 

Happy you've cleaned up the fluff!

PR #684, opened 2022-12-21
Fix #625

### LINKED ISSUE #625: A player that is out cannot collide with other players

Luxwor:
Me and Claudio fixed #625 :)

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -996,9 +996,7 @@ void test_game() //!OCLINT tests may be many
     assert(is_dead(p));
   }
 #endif
-  #define FIX_ISSUE_625
-  #ifdef FIX_ISSUE_625
-    {
+  {
     //(625) A player that is dead cannot collide with other players
     game g;
     const coordinate c_p2 = g.get_player(1).get_position();
@@ -1006,8 +1004,7 @@ void test_game() //!OCLINT tests may be many
     assert(are_colliding(g.get_player(0), g.get_player(1)));
     g.kill_player(0);
     assert(!are_colliding(g.get_player(0), g.get_player(1)));
-    }
-  #endif
+  }
 
   // (236) When a player touches food it eats it
   {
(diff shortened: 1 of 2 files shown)


src/game.cpp
```diff
@@ -996,10 +996,10 @@ void test_game() //!OCLINT tests may be many
     assert(is_dead(p));
   }
 #endif
-
+  #define FIX_ISSUE_625
```
richelbilderbeek, line 999: Well done, you can remove the `#defines` after fixing an Issue :-)

src/player.cpp
```diff
@@ -129,11 +129,16 @@ void add_action(player& p, action_type action) noexcept
 
 bool are_colliding(const player &lhs, const player &rhs) noexcept
 {
+  if (lhs.get_state() == player_state::active &&
+      rhs.get_state() == player_state::active)
+  {
```
richelbilderbeek, line 134: In C++ it is common to do

```c++
if (a == b
  && c == d
  && e == f
)
{
  // Stuff
}
```

Notice the place for the `&&`: it's at the **start** at each line (note in R one cannot do this)

richelbilderbeek (changes requested):
Nearly there :-)

src/player.cpp
Luxwor: did it!

richelbilderbeek (approved):
Brilliant!

PR #683, opened 2022-12-20
Tests for Milestone 5

TheoPannetier:
[Milestone 5: ](https://github.com/tresinformal/drakkar/milestone/5)The game eventually ends and decides on a winner.

**Tests for:**
- #681 
- #682

PR #686, opened 2022-12-21
#679 rm all mentions of food

### LINKED ISSUE #679: Remove food from the game
### Context

The minimum viable game plan excludes food items; therefore, they should be removed until we have at least a `v1.0`.

EvoLandEco:
fixes #679

### DIFF
--- a/game.pri
+++ b/game.pri
@@ -6,9 +6,6 @@ HEADERS += \
     $$PWD/src/coordinate.h \
     $$PWD/src/environment.h \
     $$PWD/src/environment_type.h \
-    $$PWD/src/food.h \
-    $$PWD/src/food_state.h \
-    $$PWD/src/food_type.h \
     $$PWD/src/game.h \
     $$PWD/src/game_functions.h \
     $$PWD/src/game_options.h \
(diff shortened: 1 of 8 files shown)


EvoLandEco:
check failed:
```
Run make debug --silent
error: #endif without #if
 1336 | #endif // no tests in release
      |  ^~~~~
make[1]: *** [Makefile.Debug:463: debug/game.o] Error 1
make: *** [Makefile:56: debug] Error 2
Error: Process completed with exit code 2.
```

but there is not line 1336 in file game.cpp, what could be wrong here? There is no warning no error on my machine

TheoPannetier:
> but there is no line 1336 in file game.cpp, what could be wrong here? There is no warning no error on my machine

The error message says that the final closing `#endif` directive has not been matched by a corresponding opening `#ifdef` or `#ifndef`, while we know it should be matched by the first `#ifndef NDEBUG // no tests in release` (on line 365 or so). This implies that another `#ifdef` earlier has lost its own matching `#endif`.

TheoPannetier:
I have investigated a bit and found two places with dangling processives. One of them is recent my fault I believe (comes from #683), apologies! The other is older, in the middle of the tests. I have fixed those, let's see if it fixes the build.
Also, it's odd that building on GHA catches this issue while our local builds do not.

TheoPannetier (approved):
My fix worked, so I'll proceed with reviewing (and merging) this.
Good work, thanks for taking the tedious task of removing all of the food mentions :)

EvoLandEco:
@TheoPannetier Great work! Thanks for the fix

PR #689, opened 2022-12-23
End game condition check

### LINKED ISSUE #682: A game is over when the time limit is reached
### Depends

- [ ] #681

### Context

The game needs to be able to check itself for when it is time to finish. The game should terminate when the game hits the time limit.

### Test

```cpp
//#define FIX_ISSUE_682
    #ifdef FIX_ISSUE_682
    // (682) A game is over when the time limit is exceeded
    {
      const int time_limit = 10;
      const game_options g_options{3,
                            false,
                            get_random_kam(),
                            get_random_kam(),
                            get_random_kam(),
                            environment_type(),
                            time_limit
                           };
      game g{g_options};
      for (int i = 0; i < time_limit; i++)
        {
          assert(!g.is_over());
          g.tick();
        }
      assert(g.is_over());
    }
    #endif // FIX_ISSUE_682
```

EvoLandEco:
fixes #682

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -223,7 +223,7 @@ void game::reset_player_action()
 /// END GAME CONDITION ///
 void game::check_over()
 {
-  if (m_n_ticks >= m_options.get_max_time()) m_is_over = true;
+  m_is_over = m_n_ticks >= m_options.get_max_time();
 }
 /// END GAME CONDITION ///
 
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -159,7 +159,7 @@ void game::tick()
   // and updates m_n_ticks
   increment_n_ticks();
 
-  // Check whether end game contion is met
+  // Check whether end game condition is met
   check_over();
 }
 
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -158,6 +158,9 @@ void game::tick()
 
   // and updates m_n_ticks
   increment_n_ticks();
+
+  // Check whether end game contion is met
+  check_over();
 }
 
 void game::increment_n_ticks()
@@ -216,6 +219,13 @@ void game::reset_player_action()
       p.set_action_flag(action_type::idle);
     }
 }
+
+/// END GAME CONDITION ///
+void game::check_over()
+{
+  if (m_n_ticks >= m_options.get_max_time()) m_is_over = true;
+}
+/// END GAME CONDITION ///
 
 double calc_mean(const std::vector<double>& v)
 {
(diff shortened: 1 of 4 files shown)


src/game_options.cpp
```diff
@@ -188,7 +190,7 @@ void test_game_options()
           environment_type(),
           max_time
           );
-    assert(game_options.get_max_time() == max_time);
+    assert(random_game_options.get_max_time() == max_time);
```
TheoPannetier, line 193: Oops, thanks for catching this mistake in the test!

src/game_options.cpp
```diff
@@ -170,11 +172,11 @@ void test_game_options()
     game_options go;
     assert(go.get_environment_type() == environment_type::empty);
   }
-//#define FIX_ISSUE_681
+#define FIX_ISSUE_681
 #ifdef FIX_ISSUE_681
   // (681) The game_options set a time limit
   {
-    const int max_time = 500; // change if too long/short
+    const int max_time = 6000; // change if too long/short
```
TheoPannetier, line 179: That looks appropriate :)

src/game.cpp
```diff
@@ -1172,8 +1182,29 @@ void test_game() //!OCLINT tests may be many
...
+    game g{g_options};
+    for (int i = 0; i < time_limit; i++)
+      {
+        assert(!g.is_over());
+        g.increment_n_ticks();
+        g.check_over();
+      }
+    assert(g.is_over());
+  }
+
```
TheoPannetier, line 1207: I feel this test block is redundant with the next one. We don't need to explicitly test that `check_over()` exists: its existence (or something similar to it) is anyways required to make the test block below pass.

```suggestion
```

src/game.cpp
```diff
@@ -217,6 +220,13 @@ void game::reset_player_action()
     }
 }
 
+/// END GAME CONDITION ///
+void game::check_over()
+{
+  if (m_n_ticks >= m_options.get_max_time()) m_is_over = true;
```
TheoPannetier, line 226: This is good, but you could make it even more concise if you like ;)

TheoPannetier (approved):
Excellent solution! I've made a suggestion to remove the extra test you have added (I feel it's redundant with the existing one), but feel free to disagree and merge as is.

EvoLandEco:
> Excellent solution! I've made a suggestion to remove the extra test you have added (I feel it's redundant with the existing one), but feel free to disagree and merge as is.

I was concerning that we might change the way we implement things later, an example is that we either implement a function outsides a class, or as a public method, I think it would be even safer to have the test to ensure that we do not accidentally change the exposed API

src/game_options.cpp
EvoLandEco: Might still be too short for one round, but that's not a thing we should think about for now

Issue #692, opened 2022-12-23
Show player size on the fourth panel


EvoLandEco:
Dependent

- [x] #690
- [x] #691

A consensus (I assume) is that the winning condition of the game is size-related, for debugging purposes we should have the players' sizes shown on the fourth panel, below the players' speeds

EvoLandEco:
Closed by #693

Issue #691, opened 2022-12-23
Rename `game_view::set_player_coords_view` to `game_view::set_player_info_view`


EvoLandEco:
the same reason as #690 , if I understand this function correctly, it sets up the fourth panel to draw player info, thus renaming it.

EvoLandEco:
Closed by #693

PR #693, opened 2022-12-23
Show player size on the fourth panel

### LINKED ISSUE #690: Rename `game_view::draw_player_coords` to `game_view::draw_player_info`
This function currently does more than its name indicates, I plan to make this function show more info (for example player size) on the fourth panel, let's rename it to `game_view::draw_player_info`

EvoLandEco:
fixes 

#690 
#691 
#692

### DIFF
--- a/src/game_view.cpp
+++ b/src/game_view.cpp
@@ -214,7 +214,7 @@ void game_view::set_player_coords_view() noexcept
     m_window.setView(player_coords_view);
 }
 
-void game_view::draw_player_coords() noexcept
+void game_view::draw_player_info() noexcept
 {
     sf::Text text;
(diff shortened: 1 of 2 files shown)


src/game_view.cpp
```diff
@@ -204,7 +203,7 @@ void game_view::draw_players() noexcept //!OCLINT too long indeed, please
     }
 }
 
-void game_view::set_player_coords_view() noexcept
+void game_view::set_player_info_view() noexcept
```
TheoPannetier, line 206: Definitely a good idea to refactor this into a name that is more general and likely to match future changes in the function!

TheoPannetier (approved):
Looks all good, thanks for making debugging easier with this extra info!

Issue #469, opened 2022-01-20
Players have an invulnerability duration


TheoPannetier:

EvoLandEco:
@ollyturner what about adding this to our collection too 🚀?

ollyturner:
Good Idea @EvoLandEco. I will add it

Issue #382, opened 2021-09-22
Invulnerable players cannot shrink [WIP]


Clemcarle:
#ifdef FIX_ISSUE_382
  ///An invulnerable player cannot shrink
  {
    game g;
    
    //Make the first player invulnerable
    become_invulnerable(g.get_player(0));
    
    // Make player 1 and 2 overlap
    g.get_player(1).set_x(g.get_player(0).get_x());
    g.get_player(1).set_y(g.get_player(0).get_y());
    assert(has_collision(g));
    
    // After a tick, invulnerable player does not shrink
    const int inv_player_size_before =  get_nth_player_size(g, 0);
    g.tick();
    const int inv_player_size_after =  get_nth_player_size(g, 0);
    assert(inv_player_size_after == inv_player_size_before);
  }
#endif

EvoLandEco:
An invulnerable player should have no clip with others, thus no grow either

Issue #381, opened 2021-09-22
Create "invulnerable" status [WIP]


Clemcarle:
Right now, a player can be shrunk many times in quick succession. We want there to be a short period of time after being shrunk where the player is not affected by shrinking. Create an "invulnerable" player state. 

```
#ifdef FIX_ISSUE_378
  ///A player can become invulnerable
  {
    game g;
    
    //Make the first player invulnerable
    assert(is_active(g.get_player(0));
    become_invulnerable(g.get_player(0));
    assert(is_invulnerable(g.get_player(0)));
    
    // Make player 1 and 2 overlap
    g.get_player(1).set_x(g.get_player(0).get_x());
    g.get_player(1).set_y(g.get_player(0).get_y());
    assert(has_collision(g));
    
    // After a tick, invulnerable player does not shrink
    const int inv_player_size_before =  get_nth_player_size(g, 0);
    g.tick();
    const int inv_player_size_after =  get_nth_player_size(g, 0);
    assert(inv_player_size_after = inv_player_size_before);
  }
#endif
```

EvoLandEco:
This issue is now related to the game play improvements, I've labelled this

TheoPannetier:
@EvoLandEco what does this issue depends on?

EvoLandEco:
@TheoPannetier We plan to implement `player_attributes` class later, to allow for player-specific invulnerable time. So I think it will depend on that later, and I marked this issue to ensure we include this

PR #695, opened 2022-12-23
Implement invulnerability after shrinking

### LINKED ISSUE #535: [Game Play] Too much 'grow' and 'shrink'
Players become very large when it collides with other players. The same is true for the opposite, as the other player shrinks greatly.
We can adjust the effects of 'grow' and 'shrink', to a reasonable level

### LINKED ISSUE #699: Expand `player_state`
Currently, `player_state` has the following members:
```c++
enum class player_state
{
  active,
  dead
};
```
My proposed change:
```c++
enum class player_state
{
  active, // a normal state
  passive, // a state that the player is non-interactable
  dead // deactivated
};
```
A passive state is that a player is invulnerable **_and_** cannot interact with the other players/enviroment, we should distinguish this from the real invulnerable state. When a player shrinks, it should become passive, not only invulnerable.

EvoLandEco:
fixes #535 
The logic is:

1. When a player loses, it shrinks and becomes passive for a while
2. When a player is passive, it cannot grow or shrink, it cannot make other players grow or shrink either, it will be like a ghost
3. After a while, an passive player will become active again


TheoPannetier:
I've assigned you to #381 #382 #463; please make sure that you assign yourself to issues before you start working on them! Otherwise other people may decide to work on it too, come up with a different solution, eventually causing a merge conflict (and some wasted time)!

EvoLandEco:
> I've assigned you to #381 #382 #463; please make sure that you assign yourself to issues before you start working on them! Otherwise other people may decide to work on it too, come up with a different solution, eventually causing a merge conflict (and some wasted time)!

Yes, I should have done it before this PR. I also assigned myself #469 😸

EvoLandEco:
When I was thinking about adding visual effect for invulnerability, I feel like the current implementation is bad. The reason is, we already have a `player_state` class, a better option would be to expand this enum class, add some states into it, and build the whole thing on top of it, rather than inventing a new flag `m_invulnerability` in the `player` class.

@TheoPannetier @swom apologies for inviting you as reviewers prematurely and (maybe) have already wasted a bit of your time. I will re-implement invulnerability,

src/player.cpp
```diff
@@ -129,8 +129,8 @@ void add_action(player& p, action_type action) noexcept
 
 bool are_colliding(const player &lhs, const player &rhs) noexcept
 {
-  if (lhs.get_state() == player_state::active
-     && rhs.get_state() == player_state::active)
```
EvoLandEco, line 133: Things become more complicated here: how do we define colliding? Simply two players are overlapping? Or they should all be non-passive and/or non-invulnerable? Here I chose to define colliding as they are simply overlapping, so as long as they are not out/dead, they can collide. This will not break the other part of the code. We might need to discuss those mechanics in depth later. 

We could add a feature later to make non-passive players unable to overlap with each other,

TheoPannetier: Yeah I think it confused me too the first time I saw this; colliding means overlapping AND active, i.e. capable to interact. If two players (or other objects) are overlapping but one or either is not active, there is no collision. I can't think of a better name for this, unfortunately!

src/game.cpp
```diff
@@ -578,54 +629,193 @@ void test_game() //!OCLINT tests may be many
...
+  /// A player can become non-passive again, using player.remove_state_passive()
+  {
+    game g;
 
+    assert(is_active(g.get_player(0)));
+    g.get_player(0).set_state_passive();
+    g.get_player(0).remove_state_passive();
+    assert(!is_passive(g.get_player(0)));
   }
   #endif
```
TheoPannetier, line 671: I don't understand why you changed the tests here. I see that you're being more exhaustive, systematically asserting that all setters exist. I don't see the necessity for this; the short test I had written was enough to reach the goal of the issue ("A player can become passive/invulnerable"). Being able to turn it back is not a necessity to reach this.
I believe (I'm not sure, can't read it from your commit history) what you are doing here is characteristic of unit-testing, that is writing tests, typically after the code, to ensure that the concerned bit of code keeps working as intended. This is ok, but we prefer test-driven development. The [difference ](https://methodpoet.com/unit-testing-vs-tdd/) is that in TDD, we write the test *before* the code and even before thinking of the implementation. The test states our intended goal, and will pass when this goal is reached. Then we seek to implement a solution.

This is good work, your tests are functional and do satisfy the end goal, I just want to raise the point that I believe this is excessive, maybe premature testing.

src/game.cpp
```diff
@@ -578,54 +629,193 @@ void test_game() //!OCLINT tests may be many
...
 
+  #define FIX_ISSUE_382
   #ifdef FIX_ISSUE_382
-  ///An invulnerable player cannot shrink
+  /// An passive player cannot grow
```
TheoPannetier, line 675:
```suggestion
  /// A passive player cannot grow
```

src/game.cpp
```diff
@@ -578,54 +629,193 @@ void test_game() //!OCLINT tests may be many
...
+
+    // After a tick, passive player does not grow
+    const int inv_player_size_before =  get_nth_player_size(g, winner_index);
     g.tick();
-    const int inv_player_size_after =  get_nth_player_size(g, 0);
+    const int inv_player_size_after =  get_nth_player_size(g, winner_index);
+    assert(inv_player_size_after == inv_player_size_before);
+  }
+
+  /// An passive player cannot shrink
```
TheoPannetier, line 699:
```suggestion
  /// A passive player cannot shrink
```

src/game.cpp
```diff
@@ -578,54 +629,193 @@ void test_game() //!OCLINT tests may be many
...
     assert(inv_player_size_after == inv_player_size_before);
   }
   #endif
 
-  //#define FIX_ISSUE_463
+
+  #define FIX_ISSUE_463
   #ifdef FIX_ISSUE_463
-  // Players lose invulnerability after a short time
+  // Can get and set a player's passive state
```
TheoPannetier, line 727:
```suggestion
  // Can get and set a player's passive state duration
```

src/game.cpp
```diff
@@ -578,54 +629,193 @@ void test_game() //!OCLINT tests may be many
...
+    game g;
+    player& p = g.get_player(0);
+
+    // Check default duration
+    assert(p.get_duration_passive() == 2000);
+
+    int new_duration = 200;
+    p.set_duration_passive(new_duration);
+    assert(p.get_duration_passive() == new_duration);
+  }
```
TheoPannetier, line 738: I can't see why you would change the duration of the passive state at the player level; shouldn't it be a fixed parameter of the game?

src/game.cpp
```diff
@@ -578,54 +629,193 @@ void test_game() //!OCLINT tests may be many
...
+    g.increment_passive_timers();
+    assert(p.get_passive_timer() == 0);
+
+    int timer_before = p.get_passive_timer();
+    p.set_state_passive();
+    g.increment_passive_timers();
+    int timer_after = p.get_passive_timer();
+    int difference = timer_after - timer_before;
+    assert(difference == 1);
+  }
```
TheoPannetier, line 756: Same comment as above for test 381; I don't think this adds anything that is not already (implicitly) required by the test I had written (below)

src/game.cpp
```diff
@@ -578,54 +629,193 @@ void test_game() //!OCLINT tests may be many
...
+      int timer_after = p.get_passive_timer();
+      int difference = timer_after - timer_before;
+      assert(difference == 1);
+      assert(timer_after <= duration);
+      timer_before = p.get_passive_timer();
+      g.reset_passive_timers();
+      timer_after = p.get_passive_timer();
+      assert(timer_after == timer_before);
+    }
+  }
```
TheoPannetier, line 805: Same as above, I don't think it's necessary to explicitly test for the timer?
Unless you think the issue needed to be broken down in smaller steps - in which case, sure, go ahead and create smaller intermediate issues towards losing invulnerability after some time :)

src/game.cpp
```diff
@@ -294,7 +336,12 @@ void grow_winning_player(game &g)
 
   const int winner_index = get_winning_player_index(g, first_player_index, second_player_index);
   player& winning_player = g.get_player(winner_index);
-  winning_player.grow();
+  const int loser_index = get_losing_player_index(g, first_player_index, second_player_index);
+  player& losing_player = g.get_player(loser_index);
+  if (!is_passive(winning_player) && !is_passive(losing_player))
```
TheoPannetier, line 341: I think passive states shouldn't be caught here; instead if either player is passive, there is no collision (and thus, growing/shrinking won't be called)

src/game.cpp
```diff
@@ -143,6 +143,12 @@ void game::tick()
     shrink_losing_player(*this);
   }
 
+  /// PASSIVE STATE ///
+  increment_passive_timers();
```
TheoPannetier, line 147: I disagree with the refactoring of `invulnerable` into `passive`. I think the former conveys more explicit information than the latter. `passive` is too general, could refer to the player being inactive for any reason, e.g. having been killed, stunned, or recovering from a defeat, and suggest the player cannot interact with anything, including foods etc. `invulnerable` simply means the player cannot be attacked, and lose a fight

TheoPannetier (changes requested):
I have been a bit harsh and grumpy in my comments, but overall this is good work and this contribution definitely helps the game move forward :)
So there is only one point where I request a change before merging: collision should not happen if either member is passive, and in turn there is no need to check for state in grow and shrink.
I disagree with the name passive instead of invulnerable, but the last word is yours for this :+1:

src/game.cpp
EvoLandEco: @TheoPannetier Thanks for pointing out this, it was that I feel like the tests I had written beforehand were not enough to detect all the changes in the future that might be done intentionally or unintentionally by me or by others, I am sometimes too obsessive that I force myself to add a lot of probably excessive or premature tests. 😢

src/game.cpp
EvoLandEco: @TheoPannetier I was thinking about possible future extension of the power-ups etc. For a minimal viable game, it should also be good to just let the duration be a fixed parameter of the game. What is your opinion about this?

src/game.cpp
EvoLandEco: I added these tests for debugging purposes as I encountered quite a lot of issues. I was making sure that all the parts I had added were working as I expected.

src/game.cpp
EvoLandEco: IMO we do need to have some sound definitions of these states. For the `passive` state, I was referring to a game state that is the same as in GTA5, which means a player is alive but cannot be killed by the others players, and of course, a passive player cannot hurt the other players. A passive player can collect things and do jobs.

Considering foods or power-ups, it is worthwhile to think thoroughly. Maybe we could find a more proper name for this state I was referring to? As for `invulnerable`, I still feel like it could be a state after collecting a power-up or triggering some game mechanics.

EvoLandEco:
@TheoPannetier An open discussion is always helpful, thanks for the opinions, and let's keep talking and find a best solution 😉

src/game.cpp
TheoPannetier: >  I am sometimes too obsessive

I definitely can't blame you for *caring* about the future of the game! I confirm I was too harsh with my initial comment, and I'd like to apologise for this. 

These tests are indeed a bit premature at this stage, but they don't remove anything or add anything dangerous to the game AFAICS and may pave the way for the next steps with the passive state. Take the above as advice on how to better design tests, but what you're doing is an important step too.

src/game.cpp
TheoPannetier: I think you've figured out my opinion already 😉 
I believe anticipating possible extensions beyond the minimal viable game is premature optimisation, and a wise man once told me it's the [root of all evil](https://effectiviology.com/premature-optimization/).

So for now, I'd rather have the duration as a fixed parameter. It will be easy to make it free again when/if we need it :)

src/game.cpp
TheoPannetier: The main test (if it's good!) should already ensure that everything works as expected. If players lose invulnerability after some time and not earlier, as #463 demands, then we have our feature and we're happy!
I'd argue we only care about the underlying mechanisms and potential issues *if* a problem indeed manifests after completing the test. Then it's time to make a new issue to signal the problem, and write a test that ensures it is fixed.

That being said, using tests as debug towards completing another test is entirely fine, but then no need to leave it in the code once you have resolved the final test!

src/game.cpp
TheoPannetier: Fair enough, I agree with the reasoning, although again I prefer `invulnerable` over `passive` to refer to what you describe: `passive` implies you cannot do anything other than move, so no item or task collection. On the other hand, `invulnerable` does suggest you could still attack other players, which is not true (or is it?). 
I'll let you make the final call :)

src/game.cpp
EvoLandEco: > The main test (if it's good!) should already ensure that everything works as expected. If players lose invulnerability after some time and not earlier, as #463 demands, then we have our feature and we're happy! I'd argue we only care about the underlying mechanisms and potential issues _if_ a problem indeed manifests after completing the test. Then it's time to make a new issue to signal the problem, and write a test that ensures it is fixed.
> 
> That being said, using tests as debug towards completing another test is entirely fine, but then no need to leave it in the code once you have resolved the final test!

Yes that's true, I mean I would add extra debugging tests besides the main tests and remove them right before merging, this could be a better way than what was done in this PR 😉

src/game.cpp
EvoLandEco: > Fair enough, I agree with the reasoning, although again I prefer `invulnerable` over `passive` to refer to what you describe: `passive` implies you cannot do anything other than move, so no item or task collection. On the other hand, `invulnerable` does suggest you could still attack other players, which is not true (or is it?). I'll let you make the final call :)

I think what is important is to have two types of invulnerable (not receiving damage) states, one you can still deal damage while the other one you cannot.

To be more specific:

- Invulnerable State A: If a player receives damage, it shrinks and becomes State A, in this state it won't receive more damage and won't deal damage or won't collide, it is like `passive` we've discussed.
- Invulnerable State B: If a player collects a power-up or something, it just won't receive any damage, but it still deals damage to the others.

Is there any possibility we find two words to well describe them?

src/game.cpp
EvoLandEco: > > I am sometimes too obsessive
> 
> I definitely can't blame you for _caring_ about the future of the game! I confirm I was too harsh with my initial comment, and I'd like to apologise for this.
> 
> These tests are indeed a bit premature at this stage, but they don't remove anything or add anything dangerous to the game AFAICS and may pave the way for the next steps with the passive state. Take the above as advice on how to better design tests, but what you're doing is an important step too.

This mechanism really involves a lot of worth-thinking issues, I've been a bit busy with my paper recently, I would find a time to rethink about it, and try to improve this PR

src/game.cpp
EvoLandEco: > I think you've figured out my opinion already 😉 I believe anticipating possible extensions beyond the minimal viable game is premature optimisation, and a wise man once told me it's the [root of all evil](https://effectiviology.com/premature-optimization/).
> 
> So for now, I'd rather have the duration as a fixed parameter. It will be easy to make it free again when/if we need it :)

I'll change it in the following commits

src/game.cpp
TheoPannetier: > Invulnerable State B: If a player collects a power-up or something, it just won't receive any damage, but it still deals damage to the others

We don't have power-ups are plans to add any before `1.0`, so let's not worry about it for now ;)
We can always make changes later if the current implementation becomes a barrier to new desired features.
Let's keep it to what's referred to in the title of this PR.
So for this "invulnerable state A", I guess either `passive` or `invulnerable` are ok names; neither is entirely satisfying but they do an ok job.

TheoPannetier:
Hello @EvoLandEco !
I think the discussion has been a bit dense and dispersed, so to make your life easier here are the two things I'd like to see addressed before I accept this PR:

1. When resolving collision and player growth/shrinkage, you check that players involved in the collision are both active before resolving the effects of the collision. This is good, but you do the check inside the growing/shrinking functions. I feel that if either player is not active, there should not be any collision at all (and no need to call the growth/shrinkage functions). So the check for player active/passive state should be done earlier, and `has_any_player_collision` should return FALSE if either player is passive.
2. Make passive duration a fixed parameter of either the `player` or `game`, and remove the corresponding setter.

(and as usual, resolve any pending merge conflict!)

src/game.cpp
EvoLandEco: @TheoPannetier That's true, I will refactor `passive` back to `invulnerable` for now

EvoLandEco:
> Hello @EvoLandEco ! I think the discussion has been a bit dense and dispersed, so to make your life easier here are the two things I'd like to see addressed before I accept this PR:
> 
> 1. When resolving collision and player growth/shrinkage, you check that players involved in the collision are both active before resolving the effects of the collision. This is good, but you do the check inside the growing/shrinking functions. I feel that if either player is not active, there should not be any collision at all (and no need to call the growth/shrinkage functions). So the check for player active/passive state should be done earlier, and `has_any_player_collision` should return FALSE if either player is passive.
> 2. Make passive duration a fixed parameter of either the `player` or `game`, and remove the corresponding setter.
> 
> (and as usual, resolve any pending merge conflict!)

Let's focus on what this PR is meant to be. 😉 I remember I did status check in a later stage due to a limitation from how collision check was implemented, but I will surely revisit this part and see if I can do any better.

For the passive duration, I agree to make it align with what we did in the shoot cooldown system: maybe make it a `static` variable.

EvoLandEco:
> Hello @EvoLandEco ! I think the discussion has been a bit dense and dispersed, so to make your life easier here are the two things I'd like to see addressed before I accept this PR:
> 
> 1. When resolving collision and player growth/shrinkage, you check that players involved in the collision are both active before resolving the effects of the collision. This is good, but you do the check inside the growing/shrinking functions. I feel that if either player is not active, there should not be any collision at all (and no need to call the growth/shrinkage functions). So the check for player active/passive state should be done earlier, and `has_any_player_collision` should return FALSE if either player is passive.
> 2. Make passive duration a fixed parameter of either the `player` or `game`, and remove the corresponding setter.
> 
> (and as usual, resolve any pending merge conflict!)

@TheoPannetier We have the `game_options` class here, what about putting passive duration in this class?

TheoPannetier:
> @TheoPannetier We have the game_options class here, what about putting passive duration in this class?

Indeed, it looks appropriate :)
(sorry, only seeing this now)

TheoPannetier:
Merged to not lose progress

PR #711, opened 2023-02-01
Can run menu without sound

janclod:


TheoPannetier (approved):
Nice! I didn't know it was possible to iterate through arguments!

PR #712, opened 2023-02-01
The menu has cool background and we save on computer resource when loading file

janclod:
We load file directly from memory without copying it. I would suggest refactor the whole resource handling like this.

```
    QResource artWork1 { ":/276813041_drakkar__dragon__realistic__flying__paper__rock__scissors__RGB__3.png" };
    m_artwork_1.loadFromMemory(artWork1.data(), artWork1.size());
```

Probably safe to load from memory according to [this](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Texture.php#a2c4adb19dd4cbee0a588eeb85e52a249).

### DIFF
--- a/resources/game_resources.qrc
+++ b/resources/game_resources.qrc
@@ -12,5 +12,8 @@
         <file>marjon_the_dragon.png</file>
         <file>grow.ogg</file>
         <file>shrink.ogg</file>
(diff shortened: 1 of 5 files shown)


TheoPannetier (approved):
Good work, looking forward to see the artwork displayed on the menu background!

Issue #710, opened 2023-01-25
The music on/off button changes color dynamically


TheoPannetier:
### Context
As we're not sure whether the turning music on/off feature works, it would be great to have a visual cue.
When is on, button should be green.
When it is off, button should be red.

Instead of green/red, let's use the following proxies so that color blind people are not impeded:
Red: (219 67 37)
Green: (87 196 173)

### Test

```cpp
#define FIX_ISSUE_710
{
  // Colorblind-friendly green and red
    const color color_on{87, 196, 173};
    const color color_off{219, 67, 37};

    // Music is first off, then turned on
    const game_options go_music_off(0, false);
    options_view ov(go_music_off);
    color mb_color = ov.get_music_button().get_color();
    assert(mb_color == color_off);
    ov.click_play_music_button();
    mb_color = ov.get_music_button().get_color();
    assert(mb_color == color_on);
}
#endif // FIX_ISSUE_710
```

TheoPannetier:
Done, the music button changes color as it is turned on or off!

Issue #664, opened 2022-12-07
A `game_view`'s options can be updated


TheoPannetier:
### Context

We can change (some) Options from the Options screen (`options_view`), great!
But this currently has no effect on the options of the game run in `game_view`, because both are independent.
To solve this, we need to update the options in `game_view` with those in `options_view` every time we switch to this screen.

### Test
- [ ] Test is ready and on develop 

```cpp
// (664) A game_view's options can be updated from options_view's
  {
    game_options options(0, true);
    view_manager vm(view_mode::options, options);
    assert(vm.get_game_view().get_options().is_playing_music());

    vm.get_options_view().click_play_music_button();
    vm.set_next_view(view_mode::game);
    assert(!vm.get_game_view().get_options().is_playing_music());
  }
```

TheoPannetier:
Done

Issue #630, opened 2022-10-05
One must be able to turn off the music in the options screen


richelbilderbeek:
### Depends

- [x] #664

### Context

The options screen is a dialog in which the user can specify his/her ideal preferences for the game.
These preferences/settings are encapsulated in `game_options`.
It would make sense that the options dialog allows one to modify all `game_options`.

In this case, there should be some UI element (e.g. an on/off button) to turn on/off the sound.

This is the test:

```c++
  //#define FIX_ISSUE_630
  #ifdef FIX_ISSUE_630
  // (630) One must be able to turn off the music in the options screen
  {
    const game_options options;
    assert(options.is_playing_music());
    options_view view(options);
    assert(view.get_options().is_playing_music());
    view.click_play_music_button();
    assert(!view.get_options().is_playing_music());
  }
  #endif // FIX_ISSUE_630
```

After that, the button needs to be clickable for real. Add it and let a human test this:

 * Start the options screen, music should be on/off
 * Click twice on (e.g.) 'Music on/off' to make it turn off/on (or vice versa)

TheoPannetier:
Hey @richelbilderbeek @ollyturner looks like you guys assigned yourselves after me to this issue and came up with an alternative solution to the one I had made already, causing a merge conflict https://github.com/tresinformal/game/commit/902848c18125e60d2280ce533a0c6ff970aa6155

I gave you my blessing last week to work on my branch, but I feel leaving it in a state of merge conflict is quite ongezellig.
Also, I feel that assigning yourselves to work on an issue I already assigned myself too and work from scratch on  it undermines the whole point of assignment (that is, indicating to people you're already working on it to avoid divergences and merge conflicts).

Apologies if I misunderstood what happened, this is what I'm reading from this issue thread and the commit network.

richelbilderbeek:
Hi @TheoPannetier, you are right, thanks for calling me out on that: we indeed should have fixed your branch.

I do partially agree with overriding assignments: yes, it's good that team members can claim an Issue. OTOH, team members that are absent should not slow down the team and hence we should be flexible to some extent. As I feel that this Issue is among the top priorities, I feel it was justified to override the assignment to the Issue.

Still, we should have left the Issue in a better state than in which we found it. Sorry for that, and thanks for helping me remember that :-)

TheoPannetier:
[...]

PR #715, opened 2023-02-08
Can turn the music on/off by clicking a button in options

TheoPannetier:
Finally managed to fix 

- #630 

and its intermediate steps:

- #664 
- #710


src/game_view.h
```diff
@@ -31,13 +31,18 @@ class game_view
...
+
   /// Processes events in game and ouputs false if quit
   /// is inputted
   bool process_events();
 
   ///Gets a const ref to m_game
   const game& get_game() const noexcept {return m_game; }
 
+  ///Gets a ref to m_game
+  game& get_game() noexcept {return m_game; }
```
richelbilderbeek, line 44: Ooooh, this is scary, as one should [C.9: Minimize exposure of member](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-private): making `game` changable by ref is quite of the opposite. However, I just mention this so you can make a mental note and then observe the havoc this may cause. And hey, maybe I am wrong :innocent:

richelbilderbeek (changes requested):
Good stuff and easy to review, well done! 

I just suggest some `const`s, discuss a bit about layout and mention a getter being scary. Just read and optionally ignore :-)

I hope you'll happily fix those consts and ask for another review, after I will happily accept.

src/options_view.cpp
```diff
@@ -57,38 +80,66 @@ void options_view::show()
...
-
   // Display all shapes
   m_window.display();
 }
 
+// Click the music button to turn music on/off
+void options_view::click_play_music_button()
+{
+  if (m_options.is_playing_music()) // music is on
+    {
```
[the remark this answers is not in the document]
TheoPannetier: Ah, my bad. I prefer the correct Cpp one's as well!

src/game_view.h
TheoPannetier: Rm'd it, but in my defence, I didn't add this to the game, only changed its place 😇 
It seems that it was used for profiling.

richelbilderbeek: Ah, like that :-) . Thanks for the deletion then :grin:

PR #725, opened 2023-02-11
Towards a better menu

TheoPannetier:
Added the resources and a small starting test to start improving the graphical user interface.
See #724


richelbilderbeek (approved):
Nice! Cool that the original author is also credited!

PR #727, opened 2023-02-22
Ludvig

### LINKED ISSUE #721: The largest player wins the game
### Context

The player that wins should be the largest one.

### Test

```cpp
  #ifdef FIX_ISSUE_721
  // (721) The largest player wins the game
  {
    game g;
    player &player_two = g.get_player(1);
    player_two.grow();
    assert(g.who_is_winning() == 1);

    player &player_three = g.get_player(2);
    player_three.grow();
    player_three.grow(); // twice
    assert(g.who_is_winning() == 2);
   }
  #endif // FIX_ISSUE_721
```

Luxwor:
Fixed #721 with Claudio :)


src/game.cpp
```diff
@@ -1224,20 +1224,18 @@ void test_game() //!OCLINT tests may be many
...
   }
 
-  #ifdef FIX_ISSUE_721
   // (721) The largest player wins the game
   {
     game g;
     player &player_two = g.get_player(1);
     player_two.grow();
-    assert(g.who_is_winning() == 1);
+    assert(g.who_is_winning() == "1");
```
richelbilderbeek, line 1232: Why is this changed to a string? A player number is ... well ... a number?

richelbilderbeek: Ah, an ID is a string. Accepted!

richelbilderbeek (approved):
Good stuff, could not have done it better. An ID is a `string` :-)

Issue #722, opened 2023-02-11
In case of a tie, winner is decided on a coin flip


TheoPannetier:
### Depends

- [ ] #721

### Context

If two or more players have the same size by the end of the game, the winner should be decided on a coin flip amongst them.

### Test

```cpp
  #ifdef FIX_ISSUE_722
  // (722) In case of a tie, winner is decided on a coin flip
  {
      const int a_seed = 5;
      const int another_seed = 6; // change if this picks the same winner

      game a_game(game_options{a_seed});
      player &player_two = a_game.get_player(1);
      player_two.grow();
      player &player_three = a_game.get_player(2);
      player_three.grow();
      assert(a_game.who_is_winning() != 0);

      game another_game(game_options{another_seed});
      player &other_player_two = another_game.get_player(1);
      other_player_two.grow();
      player &other_player_three = another_game.get_player(2);
      other_player_three.grow();
      assert(another_game.who_is_winning() != 0);

      assert(a_game.who_is_winning != another_game.who_is_winning());
  }
  #endif // FIX_ISSUE_722
```

Issue #478, opened 2022-02-02
Save the game state


richelbilderbeek:
It is important to be able to save the game state, for example, in bug reports or maybe in testing, where we can *load* a game state.

richelbilderbeek:
Make this test come true:

```c++
  #ifdef FIX_ISSUE_478
  // Saving a game and loading it, must result in the same game
  {
    const game g;
    const std::string filename = "test.txt";
    save(g, filename); // To prevent a bloated/Winnebago class
    const game h = load(filename);
    assert(g == h);
  }
  #endif // FIX_ISSUE_478
```

swom:
@richelbilderbeek I think writing by hand all input and output operators is going to be quite tedious and could anyway lewad to brittle architecture. What about using the [json library](https://github.com/nlohmann/json) I use in my more recent projects [more recent projects](https://github.com/swom/ArchitectureEvolution) with just one template save and load function?

richelbilderbeek:
Hi @swom, I am quite late with responding I see. Sorry for having gotten too sloppy :-) .

I would argue **against** using another library, as this demands an action of the whole team. If we get it 'for free' using CMake, then I'd definitely be up for it! Till then, a simple CSV would suffice.

PR #729, opened 2023-03-01
Re-expose #478, update test for #722

richelbilderbeek:
The test for #478 was only partially implemented, so I resurrected it. Although save-states may not be the most important now, I am happy they are part of a test.

The test for #722 was outdated, as a return type has changed from int to string, and the final test had a missing `()`.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -844,13 +844,15 @@ void test_game() //!OCLINT tests may be many
   }
 
   // (478) Saving a game and loading it, must result in the same game
+  #ifdef FIX_ISSUE_478
   {
     const game g;
     const std::string filename = "test.txt";
     save(g, filename); // To prevent a bloated/Winnebago class
-    //const game h = load(filename);
-    //assert(g == h);
+    const game h = load(filename);
+    assert(g == h);
   }
+  #endif // FIX_ISSUE_478
 
   #define FIX_ISSUE_524
   #ifdef FIX_ISSUE_524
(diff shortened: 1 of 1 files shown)


src/game.cpp
```diff
@@ -1237,27 +1239,28 @@ void test_game() //!OCLINT tests may be many
...
       game another_game(game_options{another_seed});
       player &other_player_two = another_game.get_player(1);
       other_player_two.grow();
       player &other_player_three = another_game.get_player(2);
       other_player_three.grow();
-      assert(another_game.who_is_winning() != 0);
+      assert(another_game.who_is_winning() != "0");
 
-      assert(a_game.who_is_winning != another_game.who_is_winning());
+      assert(a_game.who_is_winning() != another_game.who_is_winning());
```
TheoPannetier, line 1263: well spotted, thanks for fixing my mistake!

src/game.cpp
```diff
@@ -844,13 +844,15 @@ void test_game() //!OCLINT tests may be many
...
   // (478) Saving a game and loading it, must result in the same game
+  #ifdef FIX_ISSUE_478
   {
     const game g;
     const std::string filename = "test.txt";
     save(g, filename); // To prevent a bloated/Winnebago class
-    //const game h = load(filename);
-    //assert(g == h);
+    const game h = load(filename);
+    assert(g == h);
```
TheoPannetier, line 853: It's always nice to see the amount of commented out code reduced!

TheoPannetier (approved):
Thanks for improving and fixing my test, and good that we have a functional test for saving the game!

Issue #728, opened 2023-03-01
Write a Design Doc


TheoPannetier:
### Context

Intentions for what should be in the game and how it should work should be gathered in a single, concise and easily-accessible document.
If we want the doc to be effective, it should be linked somewhere in the README.
Then we should only work towards creating those features that are described in the doc.

Finally, we should decide on a policy to make changes to the design doc.

TheoPannetier:
we do have a game design doc now!

Issue #726, opened 2023-02-22
There is a maximum size a player can reach


TheoPannetier:
### Context

Players are so far able to reach disproportionate sizes by continuously eating the same weaker player.
This may no longer present an issue once we have some buffer time between two successive collisions + growth (e.g. #695), but for now it would be best to cap a player's max size. Rm later on if appropriate.

### Test
- [ ] Test is ready and on develop 
^this  signals contributors that the test is ready to be solved; if not ticked a test must be written/completed first

```cpp
#define FIX_ISSUE_XXX
// Something must become true
{
  
}
#endif // FIX_ISSUE_XXX
```

TheoPannetier:
Done!

PR #730, opened 2023-03-01
Game design doc + Players have a max size

TheoPannetier:
Oops, this should have been 2 separate PRs... Oh, well.

**Game design document** (see #728 and chat in the big table)

I made a first version for a design document, detailing everything that the game should do and contain for the minimum viable game (it's not that much!). Please review and feel free to object, I may have missed something.

@EvoLandEco and @ollyturner, please pay especially close attention to the rules I detailed for player size and invulnerability. Let me know if you spot that this would not work in practice, or if you have a better system in mind, or if I have left something important aside.

Once we reach a consensus on the rules, I'll make some housekeeping in the issues to make them consistent with the proposed plan.

Also, we should have a defined policy for how to add or changes features from the design doc, when we want to take the game in a new direction. I think that any new feature or idea should be discussed first, then added to the design doc if well received, and then only we should be able to start working on it (by making issues and tests, and solving them).
Any idea on how to implement this?

---------------------------

**Test and solution for:**

- #726


README.md
```diff
@@ -52,20 +52,11 @@ contact [Stefano 'swom' Tiso](https://github.com/swom).
...
-------------------------------------------------------|------------------------
-[221](https://github.com/tresinformal/drakkar/issues/221)|-1. Rock-Paper-Scissor dynamics
-[219](https://github.com/tresinformal/drakkar/issues/219)|0. Food
-[208](https://github.com/tresinformal/drakkar/issues/208)|1. Stun
-[209](https://github.com/tresinformal/drakkar/issues/209)|2. Poison
-[210](https://github.com/tresinformal/drakkar/issues/210)|3. Regenerating walls
-[212](https://github.com/tresinformal/drakkar/issues/212)|4. Heal
-[222](https://github.com/tresinformal/drakkar/issues/222)|5. Targeting system
-[223](https://github.com/tresinformal/drakkar/issues/223)|6. Minimap
+Drakkar is a multiplayer game featuring dragons of different colours trying to chase and eat one another. The game resembles agar.io, with the addition of a Rock-Paper-Scissor dynamic via the RGB colours.
```
EvoLandEco, line 55: It makes the core idea of the game really clear

doc/game_design.md
```diff
@@ -0,0 +1,48 @@
...
+The goals and rules of the game are described here. We aim to make the design described in this doc come to life, and should refrain from working on things that do not fall in its scope. Any new feature or idea should be discussed first, then if received positively should be included in the design doc. 
+
+## General Principle
+There are three players. Each player controls a disc-shaped dragon of one of three colours: red, green, and blue.
+Every colour is strong against another one and weak against the third one. 
+Players chase one another, trying to run into and eat the weaker player while evading the stronger player.
+
+### Interactions
+A player running into another one triggers a collision. 
+Upon a collision, the colour of the involved dragons is compared, and the dragon with the strongest colour wins, according to the following Rock-Paper-Scissor logic:
```
EvoLandEco, line 14: "and the dragon with the strongest colour wins"

Maybe 'stronger' is better here? Because there is actually no strongest color, they are only stronger comparing to the colors that are weaker than them

doc/game_design.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Red is strong against Green, and weak against Blue;
+- Green is strong against Blue, and weak against Red;
+- Blue is strong against Red, and weak against Green.
+
+The winning dragon then munches on the losing one, growing a bit in the process, while the loser shrinks a bit.
+The loser then turns invulnerable for a short duration, so that it has a chance to escape its victor and avoid being munched on indefinitely.
+During this time, the invulnerable dragon cannot be involved in a collision with the stronger player (it simply ignores it as it runs into it), but can still collide with the weaker player.
```
EvoLandEco, line 22: So the idea here is that invulnerable means immune to damage but still able to deal damage

doc/game_design.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Winning condition
+The game is timed; when the timer runs out, whichever player has the largest dragon wins the game.
+In the event of a tie, the winner is decided on a coin flip among the tied players.
+
+### Death and size limitations
+As described above, players change size upon interaction with other players.
+When a player falls under a certain size, it "dies". 
+While dead, the player cannot interact with either of the other players (and so ignores collisions).
+After some time, the player is revived, upon which it returns to its starting size and can interact with other players again.
```
EvoLandEco, line 32: I am curious about the fairness to return to the starting size after revival, but it is hard to say and I don't have experience playing arga.io, we will see after achieving the minimal viable game of course. 

Also, do the kill/death stats matter in determining the winner? Might be useful if there is a tie.

doc/game_design.md
```diff
@@ -0,0 +1,48 @@
...
+
+There is also an upper limit to the size a player can reach, so it does not inflate to occupy the whole environment and automatically trigger collisions with other players.
+
+Players progressively shrink over time, as an incentive to avoid players simply sitting around and waiting for the timer to run out after having reached a winning size.
```
EvoLandEco, line 36: Will a player die due to natural shrinking? Or there will be a lower limit too?

doc/game_design.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Environment
+Players can move and interact within a rectangular environment. The environment has hard boundaries: players cannot move beyond these boundaries.
+Otherwise, players can move freely, there are no obstacles other than other players within the environment.
```
EvoLandEco, line 40: If there are no obstacles, would it be impossible to lose the enemy?

doc/game_design.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Player movement
+Players can move forward by pressing a key, immediately moving in the direction that the nose of the dragon points towards.
+A player can also move backwards, going in the opposite direction to the forward one.
+To change direction, a player can pivot left or right with dedicated keys.
+
+Movement is immediate: when a movement key is pressed, the player moves accordingly; when a movement key is released, the player stops moving.
```
EvoLandEco, line 47: In my reworked movement system there is possibility to add (significant) inertia to the players, like the Mario Brothers. If you are not satisfied with the current settings we could discuss and change.

doc/game_design.md
TheoPannetier: ah, good point!

TheoPannetier, line 14:
```suggestion
Upon a collision, the colour of the involved dragons is compared, and the dragon with the stronger colour among the two wins, according to the following Rock-Paper-Scissor logic:
```

doc/game_design.md
TheoPannetier: That's a risk, yes... If both players are equally skilled at maneuvers then the chase might be long... But I think walls help a bit already. Then indeed this is something we can aim to improve for a later version.

doc/game_design.md
TheoPannetier: Hm, this has never been discussed I think. What do you reckon is best? Maybe "natural" shrinkage should stop when the player reaches the minimum size before a player dies?

doc/game_design.md
TheoPannetier: Yes, correct! If you lose a collision and then run into the player weak to your colour, you can win.

doc/game_design.md
TheoPannetier: > Also, do the kill/death stats matter in determining the winner? Might be useful if there is a tie.

They don't, so far we solve ties with a coin flip. I like the idea of using death stats instead, sounds more fair! Let's keep it in the back of our heads for now.

TheoPannetier: > I am curious about the fairness to return to the starting size after revival, but it is hard to say and I don't have experience playing arga.io, we will see after achieving the minimal viable game of course.

This is because of the "natural decay" thing. If a player reaches the minimum size and dies, then it has no way to interact and thus it cannot grow. Upon reviving, because of the decay it will immediately die again. Being revived at your starting size avoids this. Is there a logic flaw in this?

richelbilderbeek:
It is time to merge this, thanks [to the deadline posted on Discord](https://discord.com/channels/687974673983471626/687974673983471629/1083076697424924813) and copy-pasted here:

> Dear Game Design Team, among others @Oliver and  @Tianjian, please review https://github.com/tresinformal/drakkar/pull/730 which is the Game Design Document before Wednesday March 22nd 17.00. If no one has reviewed it, it is accepted as-is.

PR #731, opened 2023-03-07
Error when compiling after reinstalling OS

janclod:
Hi @richelbilderbeek 

I am creating a PR to ask this question because it seemed the easiest way for me.

I am trying to fix the following error:
image

Do you think that it is fine to push this into `develop`? Or will your and others build break? GitHub actions seems fine... see [here](https://github.com/tresinformal/drakkar/actions/runs/4347903311)

Could you build my branch on your machine? This should already give me an indication...


src/coordinate.cpp
```diff
@@ -73,8 +73,8 @@ void test_coordinate() {
       {
         c.reset_coords_to_zero();
         c.move(forward_direction, speed);
-        const double error_x = abs(get_x(c) - speed);
-        const double error_y = abs(get_y(c) - 0);
+        const double error_x = fabs(get_x(c) - speed);
+        const double error_y = fabs(get_y(c) - 0);
```
richelbilderbeek, line 77: Ah, good find! Next time, use `std::abs` instead. The reason that `abs` works is C backwards compatibility. I've created the junior Issue #734 for just that.

src/menu.cpp
```diff
@@ -78,7 +79,7 @@ void test_menu()
   // menu_view is built with values
   // for width and height of the screen
   {
-    int width = 1000;
+    int width = 2000;
```
richelbilderbeek, line 82:
```suggestion
    const int width{2000};
```

src/menu.cpp
```diff
@@ -78,7 +79,7 @@ void test_menu()
...
     int height = 1000;
```
richelbilderbeek, line 83:
```suggestion
   const  int height{1000};
```

src/menu.cpp
```diff
@@ -116,10 +117,10 @@ void test_menu()
   {
     // in the middle
 
-    int width = 1000;
+    int width = 2000;
```
richelbilderbeek, line 120:
```suggestion
    const int width{2000};
```

src/menu.cpp
```diff
@@ -116,10 +117,10 @@ void test_menu()
...
     int height = 1000;
```
richelbilderbeek, line 121:
```suggestion
    const int height{1000};
```

richelbilderbeek (approved):
Good (and small) steps forward. Well done!

Issue #732, opened 2023-03-07
Fix bug in collision detection


richelbilderbeek:
Currently, health and diameter are the same thing, just on different scales (to be precise: `diameter == 100.0 * health`).

However, when this was implemented, some collision tests failed. These were commented out by `FIX_ISSUE_732`.

Make those tests work again :-)

OggyOggelito:
732 has nothing to do with get_diameter or health?

Luxwor:
there seams to be an issue but the test doesn't fail when I run it

PR #733, opened 2023-03-07
Fix #708, expose #732

### LINKED ISSUE #708: A player's health changes along with size

richelbilderbeek:
Me and @OggyOggelito worked on #708 and now the health **is** the diameter; to be precise, the health is the diameter divided by a 100.0.

This fix, however, exposes problems in the collision detection. Fixing that is beyond the fixed Issue, hence we opened a new issue #732, to revise the collision detection and we `IFDEF`-ed away the failing tests.

Good luck reviewing :+1:


src/player.cpp
```diff
@@ -50,20 +52,17 @@ double player::get_x() const noexcept { return m_c.get_x(); }
...
+  m_health *= m_growth_factor;
 }
 
 /// Make the player shrink
 void player::shrink()
 {
-  m_diameter /= m_growth_factor;
-  if(m_diameter <= m_death_size)
+  m_health /= m_growth_factor;
+  if(this->get_diameter() <= m_death_size)
```
janclod, line 65: Remove completely diameter, use only health

richelbilderbeek: `player::get_diameter` converts the health to the diameter, that is why it is called. You suggest a free (as in: non-class non-member) function?

janclod (approved):
We respect tests so accepted :)

TheoPannetier:
I see PR accepted; I'll take the initiative to merge 👍

